PR #1986, opened 2012-11-04
Use markdown to readme and contributing files

MarceloCajueiro:
Hello!

I'm a little freak about documentation formats so I changed the README and CONTRIBUTING files to use GitHub Flavored Markdown that is amazing for this.

Take a look at the formated [README](https://github.com/MarceloCajueiro/rubinius/blob/use_markdown_to_readme_and_contributing_files/README.markdown) and [CONTRIBUTING](https://github.com/MarceloCajueiro/rubinius/blob/use_markdown_to_readme_and_contributing_files/CONTRIBUTING.markdown) files.

I think that they are a bit more legible.

What do you guys think?


jc00ke:
7f71916 :wink:

PR #2352, opened 2013-05-17
Just fixed a README.md

rtephus:
I know, there s no description and 4 commits is not a best choise for this. Thanks.


jc00ke:
7f71916 :wink:

Issue #3181, opened 2014-11-05
Rational * BigDecimal coerces into Float


fengb:
MRI:

```
irb(main):001:0> (BigDecimal(1) * Rational(1)).class
=> BigDecimal
irb(main):002:0> (BigDecimal(1) * 0.1).class
=> BigDecimal
```

RBX 2.3.0:

```
irb(main):001:0> (BigDecimal(1) * Rational(1)).class
=> Float
irb(main):002:0> (BigDecimal(1) * 0.1).class
=> Float
```

brixen:
Rubinius will continue to maintain compatibility with the current stable version of MRI to the extent possible.

The focus for Rubinius in the near term is on the following capabilities:

1. Instruction set
2. Debugger
3. Profiler
4. Just-in-time compiler
5. Concurrency
6. Garbage collector

Contributions in the form of PRs for any of the areas of focus above, or for Ruby compatibility, are appreciated.

Issue #3180, opened 2014-11-05
rand bugs with ranges


fengb:
On MRI 2.1.4:

```
irb(main):001:0> rand(1.0..1.0)
=> 1.0
irb(main):002:0> rand(1..0)
=> nil
irb(main):003:0> rand(BigDecimal(0.01, 2)..BigDecimal(1, 2))
=> #<BigDecimal:7f966c0f1180,'0.1E-1',9(36)>
```

On Rubinius 2.3.0:

```
irb(main):001:0> rand(1.0..1.0)
ArgumentError: invalid argument - 0.0
    from kernel/common/random.rb:39:in `random'
    from kernel/common/random.rb:72:in `random_range'
    from kernel/common/random.rb:36:in `random'
    from kernel/common/kernel.rb:687:in `rand'
irb(main):002:0> rand(1..0)
NoMethodError: undefined method `coerce' on nil:NilClass.
    from kernel/delta/kernel.rb:78:in `coerce (method_missing)'
    from kernel/common/random.rb:69:in `random_range'
    from kernel/common/random.rb:36:in `random'
    from kernel/common/kernel.rb:687:in `rand'
irb(main):003:0> rand(BigDecimal(0.01, 2)..BigDecimal(1, 2))
ArgumentError: invalid argument - 0.99E0
    from kernel/common/random.rb:43:in `random'
    from kernel/common/random.rb:72:in `random_range'
    from kernel/common/random.rb:36:in `random'
    from kernel/common/kernel.rb:687:in `rand'
```

brixen:
Rubinius will continue to maintain compatibility with the current stable version of MRI to the extent possible.

The focus for Rubinius in the near term is on the following capabilities:

1. Instruction set
2. Debugger
3. Profiler
4. Just-in-time compiler
5. Concurrency
6. Garbage collector

Contributions in the form of PRs for any of the areas of focus above, or for Ruby compatibility, are appreciated.

PR #3182, opened 2014-11-05
Rand fixes

fengb:
Proposed fixes for https://github.com/rubinius/rubinius/issues/3180

Does not completely fix case 3 as there is weird type coercion that goes deeper: https://github.com/rubinius/rubinius/issues/3181

### DIFF
--- a/spec/ruby/core/kernel/rand_spec.rb
+++ b/spec/ruby/core/kernel/rand_spec.rb
@@ -53,8 +53,12 @@ describe "Kernel.rand" do
     rand(0.25..0.75).should be_kind_of(Float)
   end
 
-  it 'returns nil when range is nonexistent' do
+  it "returns nil when range is backwards" do
     rand(1..0).should be_nil
+  end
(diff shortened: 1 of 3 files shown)


spec/ruby/core/kernel/rand_spec.rb
```diff
@@ -52,6 +52,19 @@
   it "returns a float for an range argument where max is < 1" do
     rand(0.25..0.75).should be_kind_of(Float)
   end
+
+  it "returns a numeric for an range argument where max is < 1" do
+    require 'bigdecimal'
```
brixen: @fengb your proposed patch adds conditionals for `is_a?(Integer)` and `respond_to?(:to_f)`. Those cases should be possible to cover with specs that neither need BigDecimal, nor an alternate Float implementation.

Am I missing something else?

brixen:
@fengb sorry it took so long to get this merged. Thanks for helping!

PR #3247, opened 2014-12-19
Search for llvm-config on OSX using brew

joelparkerhenderson:
On Mac OSX, this command fails: `ruby-install rbx`.

The reason is that the OSX LLVM does not show up as expected. 

Many OSX developers install an additional LLVM by using the `brew` package manager: `brew install llvm`.

The `brew` manager does not replace the existing system LLVM because that would cause problems with other OSX software, for example XCode. Any approach that uses `brew link` is potentially dangerous on OSX.

Therefore, this patch adds a search for llvm-config: if the system is Darwin, and the system says that `brew` is installed, and `brew` says the formula for LLVM is installed, then the code searches for the `llvm-config` file.

I expect this code will solve the `ruby-install rbx` issue for the majority of OSX brew users. 

If this code is accepted, then subsequent improvements could add logging, advice, and corner cases such as multiple brew LLVM installations.

### DIFF
--- a/configure
+++ b/configure
@@ -599,6 +599,29 @@ Unsupported language version requested: #{version}. Options are #{@supported_ver
       end
       if which
         config = File.join(which, "llvm-config")
+      elsif @darwin
+        @log.print "  Checking for brew package manager: "
+        brew = `command -v brew`.chomp
+        if $?.success?
+          @log.write "found! #{brew}"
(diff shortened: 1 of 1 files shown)


brixen:
@joelparkerhenderson thanks for working on the patch. I applied @Roman2K's because it was simple and provides an easy way to extend the conditional for other platforms if needed.

Issue #3268, opened 2015-01-04
Incompatible implementation of Process::Status


sodabrew:
This is a spin-off from eventmachine/eventmachine#551 wherein I am trying to get unit tests running clean for all Rubies that Eventmachine supports, including Rubinius.

Ruby MRI has a C implementation of Process::Status and jams instance values into the class.
http://rxr.whitequark.org/mri/ident?i=rb_last_status_set

``` C
  rb_last_status_set(int status, rb_pid_t pid)
  {
      rb_thread_t *th = GET_THREAD();
      th->last_status = rb_obj_alloc(rb_cProcessStatus);
      rb_iv_set(th->last_status, "status", INT2FIX(status));
      rb_iv_set(th->last_status, "pid", PIDT2NUM(pid));
  }
```

Eventmachine uses similar code to construct the Process::Status result for EM.system and EM.popen, which does not work with the Rubinius Process::Status https://github.com/rubinius/rubinius/blob/master/kernel/common/process.rb#L523-584

I hacked up my Eventmachine locally and got it working with this adjustment, but it fails on MRI because the (nonexistent)initialize for Process::Status wants 0 arguments.

``` C
 static VALUE t_get_subprocess_status (VALUE self, VALUE signature)
 {
        VALUE proc_status = Qnil;

        int status;
        pid_t pid;

        if (evma_get_subprocess_status (NUM2ULONG (signature), &status)) {
                if (evma_get_subprocess_pid (NUM2ULONG (signature), &pid)) {
                        proc_status = rb_obj_alloc(rb_cProcStatus);
+                       VALUE argv[2];
+                       argv[0] = INT2FIX(pid);
+                       argv[1] = INT2FIX(status);
+                       rb_obj_call_init(proc_status, 2, argv);
                        rb_iv_set(proc_status, "status", INT2FIX(status));
                        rb_iv_set(proc_status, "pid", INT2FIX(pid));
                }
        }

        return proc_status;
 }
```

Open question: should Rubinius change its definition of Process::Status to match the way MRI does it, or should Eventmachine test its Ruby engine and construct Process::Status accordingly?

YorickPeterse:
My opinion would be "Yes, we need to change our stuff". We can make the arguments optional, which should keep compatibility with existing code. @brixen any further thoughts on this?

brixen:
Where are the specs that cover this behavior?

YorickPeterse:
[...]

PR #3270, opened 2015-01-04
Changes to make Process::Status compatible with MRI

YorickPeterse:
See https://github.com/rubinius/rubinius/issues/3268 for the issue that led to these changes. Before we merge this I'd like some feedback on this, in particular abotut the renaming of `@exitstatus` to `@status`. I don't think there's a way around this, but just in case there's something better for this I'm creating a PR instead of pushing this straight to the master branch.

### DIFF
--- a/kernel/common/process.rb
+++ b/kernel/common/process.rb
@@ -522,36 +522,39 @@ module Process
 
   class Status
 
-    attr_reader :exitstatus
     attr_reader :termsig
     attr_reader :stopsig
 
-    def initialize(pid=nil, exitstatus=nil, termsig=nil, stopsig=nil)
+    def initialize(pid=nil, status=nil, termsig=nil, stopsig=nil)
       @pid = pid
-      @exitstatus = exitstatus
+      @status = status
       @termsig = termsig
       @stopsig = stopsig
     end
 
+    def exitstatus
+      return @status
+    end
+
     def to_i
-      @exitstatus
+      @status
     end
 
     def to_s
-      @exitstatus.to_s
+      @status.to_s
     end
 
     def &(num)
-      @exitstatus & num
+      @status & num
     end
 
     def ==(other)
       other = other.to_i if other.kind_of? Process::Status
-      @exitstatus == other
+      @status == other
     end
 
     def >>(num)
-      @exitstatus >> num
+      @status >> num
     end
 
     def coredump?
@@ -559,7 +562,7 @@ module Process
     end
 
     def exited?
-      @exitstatus != nil
+      @status != nil
     end
 
     def pid
@@ -576,7 +579,7 @@ module Process
 
     def success?
       if exited?
-        @exitstatus == 0
+        @status == 0
       else
         nil
       end
--- a/kernel/common/process.rb
+++ b/kernel/common/process.rb
@@ -533,7 +533,7 @@ module Process
     end
 
     def exitstatus
-      return @status
+      @status
     end
 
     def to_i
--- a/kernel/common/process.rb
+++ b/kernel/common/process.rb
@@ -526,7 +526,7 @@ module Process
     attr_reader :termsig
     attr_reader :stopsig
 
-    def initialize(pid, exitstatus, termsig=nil, stopsig=nil)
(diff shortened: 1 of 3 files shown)


YorickPeterse:
MRI eventually ends up using this code to extract the status of a `Process::Status` object: http://rxr.whitequark.org/mri/source/process.c#302

spec/ruby/core/process/initialize_spec.rb
```diff
@@ -0,0 +1,31 @@
+require File.expand_path('../../../spec_helper', __FILE__)
+
+describe 'Process::Status#initialize' do
+  it 'initializes a new Process::Status instance' do
+    Process::Status.new.should be_an_instance_of(Process::Status)
+  end
+
+  it 'initializes Process::Status with a PID' do
```
YorickPeterse: I have a question regarding these specs. Initially I wrote them as following:

```ruby
describe 'Process::Status#initialize' do
  describe 'with a PID' do
    it 'returns the PID when calling #pid' do
      Process::Status.new(42).pid.should == 42
    end
  end

  describe 'with a PID and exit status' do
    before do
      @status = Process::Status.new(42, 1)
    end

    it 'returns the PID when calling #pid' do
      @status.pid.should == 42
    end

    it 'returns the exit status when calling #exitstatus' do
      @status.exitstatus.should == 1
    end
  end

  # ...
end
```

While this breaks the specs up more nicely and make sure _all_ attributes are set, I can't help but feel that they're getting a bit verbose. Perhaps shared examples would be a solution to this?

kernel/common/process.rb
```diff
@@ -522,36 +522,39 @@ def self.coerce_rlimit_resource(resource)
...
+    def initialize(pid=nil, status=nil, termsig=nil, stopsig=nil)
       @pid = pid
-      @exitstatus = exitstatus
+      @status = status
       @termsig = termsig
       @stopsig = stopsig
     end
 
+    def exitstatus
+      return @status
```
sodabrew: Style - `return` not needed?

sodabrew:
MRI appears to store the termsig and stopsig as high bits in `status`, always masking out the other bits when returning the return code. I think this is more consistent with the Unix standards for exit status.

http://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html

http://rxr.whitequark.org/mri/source/process.c#087
``` C
087 #ifndef WIFEXITED
088 #define WIFEXITED(w)    (((w) & 0xff) == 0)
089 #endif
090 #ifndef WIFSIGNALED
091 #define WIFSIGNALED(w)  (((w) & 0x7f) > 0 && (((w) & 0x7f) < 0x7f))
092 #endif
093 #ifndef WIFSTOPPED
094 #define WIFSTOPPED(w)   (((w) & 0xff) == 0x7f)
095 #endif
096 #ifndef WEXITSTATUS
097 #define WEXITSTATUS(w)  (((w) >> 8) & 0xff)
098 #endif
099 #ifndef WTERMSIG
100 #define WTERMSIG(w)     ((w) & 0x7f)
101 #endif
102 #ifndef WSTOPSIG
103 #define WSTOPSIG        WEXITSTATUS
104 #endif
```

kernel/common/process.rb
YorickPeterse: Good point, I'm so used to always using an explicit return that it goes automatically.

sodabrew:
Oh - @YorickPeterse interesting about .to_i, it returns all of the bits, whereas the exitstatus method returns only the 8 bit exit.

spec/ruby/core/process/initialize_spec.rb
brixen: There's not a lot that would be simpler with shared specs. This type of spec (ie initialize, check initialized) is difficult to simplify. Your approach in the current version of the specs here is fine.

The purpose of the specs is to be [correct, clear, consistent, and simple](https://groups.google.com/d/msg/rubyspec/NqCC9NtfFVs/YvcqGQ54bC0J), in that order. A more verbose spec can be all of those at once.

PR #3271, opened 2015-01-05
Alternative changes to make Process::Status compatible with MRI

sodabrew:
Alternative approach to issue #3268 instead of PR #3270.

Needs specs, still a WIP, please don't merge yet, etc.

I haven't compiled this yet on OS X due to LLVM issues (worked with ruby-build, doesn't work on the command line, I'm confused). Sorry about using Travis to do the heavy lifting.

### DIFF
--- a/kernel/common/process.rb
+++ b/kernel/common/process.rb
@@ -401,9 +401,9 @@ module Process
     # wait_pid_prim returns a tuple when wait needs to communicate
     # the pid that was actually detected as stopped (since wait
     # can wait for all child pids, groups, etc)
-    status, termsig, stopsig, pid = value
+    pid, status = value
 
-    status = Process::Status.new(pid, status, termsig, stopsig)
+    status = Process::Status.new(pid, status)
     Rubinius::Mirror::Process.set_status_global status
 
     [pid, status]
@@ -521,15 +521,11 @@ module Process
   #++
 
   class Status
-
-    attr_reader :termsig
-    attr_reader :stopsig
-
-    def initialize(pid=nil, status=nil, termsig=nil, stopsig=nil)
+    # Rubinius constructs with pid, status.
+    # MRI constructs then sets the ivars.
+    def initialize(pid=nil, status=nil)
       @pid = pid
       @status = status
-      @termsig = termsig
-      @stopsig = stopsig
     end
 
     def exitstatus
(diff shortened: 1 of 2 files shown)


YorickPeterse:
Regarding LLVM, what problems are you having? Is it complaining `llvm-config` can't be found?

sodabrew:
I'm on OS X Yosemite, so there's no gcc-apple-4.2, yet vendor/*Makefiles are still showing up with CC=gcc-apple-4.2 even after I configure with `./configure --with-vendor-yaml --llvm-config llvm-config-mp-3.5  --cc /usr/bin/clang --cxx /usr/bin/clang++`

YorickPeterse:
I believe you need to install llvm via Homebrew, see https://github.com/sstephenson/ruby-build/commit/4aeeac23ab8570ee021662d9125dfe8878b093f3.

sodabrew:
I'm using LLVM via MacPorts, and it works fine to compile Rubinius itself, but the vendor libs are picking up the old compiler. I'll tinker with it later today.

About the approach in this PR: while it better matches the MRI behavior, it totally misses the point that the status code interpretation is provided by a set of MACROS that should be implementation defined. I have no idea why MRI has a copy of those macro definitions in its own codebase.

I'm not sure how much we can or should care about Process::Status.status and Process::Status.to_i matching MRI exactly (that is, having all of the status bits not just the exit code).

YorickPeterse:
In this case my question would be: is there any code that breaks because Rubinius does not provide the same macros/bitwise stuff/etc? If the answer is yes then we should match that behaviour where possible, otherwise I don't think it makes that much sense.

Either way, this PR will have to wait until the other one is merged as otherwise we'll probably end up in Git merge hell :)

brixen:
@sodabrew @YorickPeterse is there a value exposed to Ruby in MRI that would require us to use bit-fields for compatibility?

brixen:
@sodabrew I'm not opposed to your approach but could you please reapply on master, force push, so we can clearly evaluate the changes?

sodabrew:
I rebased, but that's really just academic. I no longer believe it is a good idea to hardcode the bitwise relationships in Ruby - they are defined to be C macros in the Unix spec, so they have to be in the C++ code in `vm/builtin/system.cpp` in order to enable platform portability.

Thanks for the quick work on the Process::Status initializer!

brixen:
@sodabrew thanks for working on this and reporting the issues.

Issue #3272, opened 2015-01-05
incorrect behavior with inline rescue and compound assignment operator on rbx 2.3.x and 2.4.x


ashanbrown:
The code below fails on rbx 2.4.1.  I believe it passes on rbx 2.2.10.  I found this because it is causing failures in the inherited_resources gem specs (and presumably in the behavior of the gem as well).

```
describe "inline rescue" do
  it "properly executes compound assignment operators" do
    val = 'original'
    val ||= 'new' rescue nil
    val.should == 'original'
  end
end
```

The key item here appears to be the combination of the compound assignment operator and the inline rescue.  The following will pass.

```
describe "inline rescue" do
  it "properly executes compound assignment operators" do
    val = 'original'
    val ||= begin
      "new"
    rescue nil
    end
    val.should == 'original'
  end
end
```

I've confirmed that this fails on rbx 2.3.0 and rbx 2.4.1.  I'm pretty sure it passes on rbx-2.2.10 because inherited_resources specs pass at this version.

YorickPeterse:
In what way does the first example fail? Is `val` set to `nil` in this case?

jemc:
Also, in the first example where does the `"hello"` come from?  Did the example get mangled when you were reformatting it to share?

dontfidget:
Sorry, I rewrote the examples in rspec and screwed them up.  I've fixed them now.  I'm running a git bisect and will hopefully be able to tell you the first failing commit soon.

dontfidget:
I'm having some trouble with git bisect because of compile problems.  I can tell you that it passes through a9e2cf388eb643536b9e56495cfae85339847b95 and fails at v2.3.0.

YorickPeterse:
Minial repro (extracted from the spec example, minus the spec stuff):

```ruby
val = 'original'
val ||= 'new' rescue nil

p val
```

This produces "original" on MRI and "new" on Rbx.

YorickPeterse:
Bytecode differences, using an inline rescue:

```

============= :__script__ ==============
Arguments:   0 required, 0 post, 0 total
Arity:       0
Locals:      1: val
Stack size:  6
Literals:    5: "original", "new", :StandardError, :===, :p
Lines to IP: 1: 0..5, 2: 6..18, 0: 19..23, 2: 24..45, 0: 46..48, 2: 49..51, 4: 52..61

0000:  push_literal               "original"
0002:  string_dup                 
0003:  set_local                  0    # val
[...]

PR #3273, opened 2015-01-05
add spec for inline rescue

dontfidget:
Demonstrate the failure in https://github.com/rubinius/rubinius/issues/3272

### DIFF
--- a/spec/ruby/core/thread/inline_rescue_spec.rb
+++ b/spec/ruby/core/thread/inline_rescue_spec.rb
@@ -1,0 +1,9 @@
+require File.expand_path('../../../shared/kernel/raise', __FILE__)
+
+describe "inline rescue" do
+  it "properly executes compound assignment operators" do
+    val = 'hello'
+    val ||= 'goodbye' rescue nil
+    val.should == 'hello'
+  end
+end


YorickPeterse:
A few questions/remarks:

1. Why is this spec in `spec/ruby/core/thread`? It's not related to the `Thread` class, it should probably go in `spec/ruby/language` instead, although I'm not exactly sure in what file.
2. Please remove the empty commit used to trigger a Travis build
3. The spec descriptions need to be rewritten so it's more clear what they are about. For example, the spec isn't just testing inline rescues, it's testing the combination of `X ||= Y rescue Z`. I'll leave some code comments with some examples.

spec/ruby/core/thread/inline_rescue_spec.rb
```diff
@@ -0,0 +1,9 @@
+require File.expand_path('../../../shared/kernel/raise', __FILE__)
+
+describe "inline rescue" do
```
YorickPeterse: Since at its core this spec is testing if `||=` works correctly with a `rescue`, I'd say this description should be something along the lines of "Optional assignments".

YorickPeterse: Ah, this exists: https://github.com/rubinius/rubinius/blob/master/spec/ruby/language/variables_spec.rb#L4

Since this file is already pretty big I'd suggest putting the specs in `spec/ruby/language/optional_assignments_spec.rb`, otherwise they can be added as an extra `describe` block in this existing file.

jc00ke:
Should there be a tag? Don't we know this is going to fail?

spec/ruby/core/thread/inline_rescue_spec.rb
```diff
@@ -0,0 +1,9 @@
...
+  it "properly executes compound assignment operators" do
```
YorickPeterse: This should be something along the lines of "does not overwrite a variable when using an inline rescue". The idea is to clearly state that:

1. The spec should do something that _doesn't_ overwrite a value
2. This should not occur when using an inline rescue

dontfidget:
Sorry, I didn't mean for this pull request to be taken as is.  I just wanted to trigger a travis build to show the failure.  Sorry not to be clear about that.  I'm really not familiar with rubinious or how it is organized.  Unless you really need my assistance, I hope you and other contributors can take it from here.  Thanks.

YorickPeterse:
@dontfidget Ah ok, that's perfectly fine as well. Thanks for at least looking into this :)

PR #3274, opened 2015-01-05
String#include? no longer accepts a Fixnum.

chrisseaton:
I think back in 1.8 it would turn the lower 8 bits of a Fixnum into a
String using .chr, but that doesn't seem to be the case any more, and
it's now a TypeError to pass in a Fixnum.


jemc:
Looks good to me.
Thanks for the patch and spec!

Issue #3276, opened 2015-01-07
Using Kernel#load with wrapping breaks "top-level methods"


ngollan:
``` ruby
# anon_thing.rb
def helper_method
  puts "Found it!"
end

helper_method
```

``` ruby
# test.rb
load(
  File.join(File.dirname(__FILE__), 'anon_thing.rb'),
  true
)
```

Running `test.rb` in rbx (tried `rubinius 2.4.1.n7 (2.1.0 0a3aebab 2015-01-07 3.5.0 JI) [x86_64-linux-gnu]`) results in:

``` text
An exception occurred running test.rb:

undefined local variable or method `helper_method' on an instance of Object. (NameError)

Backtrace:

  Kernel(Object)#helper_method (method_missing) at kernel/delta/kernel.rb:78
                              Object#__script__ at anon_thing.rb:5
                            Kernel(Object)#load at kernel/common/kernel.rb:497
                              Object#__script__ at test.rb:2
               Rubinius::CodeLoader#load_script at kernel/delta/code_loader.rb:66
               Rubinius::CodeLoader.load_script at kernel/delta/code_loader.rb:152
                        Rubinius::Loader#script at kernel/loader.rb:645
                          Rubinius::Loader#main at kernel/loader.rb:799
```

In MRI it works as expected. Running without the wrapping also works in rbx.

jemc:
So, looking at this, the difference here is that MRI will create a different `main` object to use when you specify `wrap=true`, while Rubinius is always using the same (constant `MAIN`).

You can see this by modifying your example as shown:
```ruby
# issue3276.rb
p [self, self.object_id]
load(
  File.join(File.dirname(__FILE__), 'issue3276_module.rb'),
  true
)
p [self, self.object_id]
```
```ruby
# issue3276_module.rb
p [self, self.object_id]
```
```ruby
# output with MRI 2.1.2
[main, 69835780739580]
[main, 69835782996080]
[main, 69835780739580]
```
```ruby
# output with rbx 2.4.1
[main, 336]
[main, 336]
[main, 336]
```

The Rubinius code that is responsible for this behavior is `CompiledCode#create_script` defined in `kernel/common/compiled_code.rb` (https://github.com/rubinius/rubinius/blob/master/kernel/common/compiled_code.rb#L184-L203).

[...]

PR #3282, opened 2015-01-13
Add failing specs for code loading with wrapping

ngollan:
Specs the expected behaviour for issue #3276 and splits up an existing spec into multiple asserts.

### DIFF
--- a/spec/ruby/core/kernel/shared/load.rb
+++ b/spec/ruby/core/kernel/shared/load.rb
@@ -81,12 +81,40 @@ describe :kernel_load, :shared => true do
     lambda { @object.load(path) }.should raise_error(LoadError)
   end
 
-  it "sets the enclosing scope to an anonymous module if passed true for 'wrap'" do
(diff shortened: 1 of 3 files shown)


spec/ruby/core/kernel/shared/load.rb
```diff
@@ -81,12 +81,40 @@
...
+      path = File.expand_path "wrap_fixture.rb", CODE_LOADING_DIR
+      @object.load(path, true)
 
-    Object.const_defined?(:LoadSpecWrap).should be_false
-    ScratchPad.recorded.first.should be_an_instance_of(Class)
+      ScratchPad.recorded.first.should be_an_instance_of(Class)
+    end
+
+    describe "with top-level methods" do
+      before :each do
```
YorickPeterse: This can just be `before do` instead.

YorickPeterse:
Merged in as dfb3f71f85083bd958ac02053938b5e568e846d7 and 5785ba2cf0ba2adbc31f7055ee6bf14a18a3e2ac, thanks! :+1:

PR #3289, opened 2015-01-19
Adding RBXOPT information to help on -X options

jasonrclark:
Would this work for surfacing a bit more info on RBXOPT? I had looked at the
--help output and found -Xhelp, but I didn't know clearly how I could pass
those options in.

### DIFF
--- a/kernel/loader.rb
+++ b/kernel/loader.rb
@@ -426,6 +426,10 @@ VM Options
      All variables, even ones that the VM doesn't understand, are available
      in Rubinius::Config.
 
+     These options can also be passed via the RBXOPT environment variable:
+
+       RBXOPT=-Xint rbx
+
      A number of Rubinius features are driven by setting these variables.
       DOC
 


jemc:
Note that there are actually three ways to pass in -X options - it would be nice to list them all here.  They are documented in the source here:

https://github.com/rubinius/rubinius/blob/master/vm/environment.cpp#L244-L253

```c++
    /* We parse -X options from three sources in the following order:
     *
     *  1. The file .rbxrc in the current working directory.
     *  2. The RBXOPT environment variable.
     *  3. The command line options.
     *
     * This order permits environment and command line ot override
     * "application" configuration. Likewise, command line can override
     * environment configuration.
     */
```

jemc:
Heh, also just noticed a typo in the source code I referenced above.
`ot override` => `to override`

jemc:
I addressed my concerns in https://github.com/rubinius/rubinius/commit/e0f904af8c5959a85901265e7c963c041e3f552c

jasonrclark:
Thanks for tightening that up @jemc! This will be a great help to anyone like me looking to set those options.

PR #3297, opened 2015-01-28
Add support for automatic configuration of llvm-config under MacPorts.

CodeGnome:
## What does this pull request do?
- Adds support for finding a the latest usable _llvm-config*_ binary under MacPorts.
    - Enable detection of MacPorts via `Configure#macports?`.
    - Enable automatic detection of a MacPorts version of _llvm-config_ when MacPorts is installed.
    - Query MacPorts for the latest usable version of _llvm-config_ via `Configure#macports_llvm_config`.
- Addresses the _llvm-config_ portion of https://github.com/postmodern/ruby-install/issues/181, although `CC=clang` must still be passed explicitly.

## How were the changes tested?
- Tested build of configuration on Yosemite with `CC=clang ./configure`:

    > Detected old configuration settings, forcing a clean build
    >  Checking for 'llvm-config': found! (version 3.3 - api: 303)
    > [...]
    > Rubinius (f85e0a30) has been configured.

- Tested new Rubinius build with with test suite using `bundle exec rake`:

    > 2162 files, 20146 examples, 153473 expectations, 0 failures, 0 errors

### DIFF
--- a/configure
+++ b/configure
@@ -600,8 +600,12 @@ Unsupported language version requested: #{version}. Options are #{@supported_ver
       if which
         config = File.join(which, "llvm-config")
       elsif @darwin
-        out = `brew list llvm | grep '/llvm-config$'`
-        config = out.chomp if $?.success?
+        if macports?
+          config = macports_llvm_config
+        else
+          out = `brew list llvm | grep '/llvm-config$'`
+          config = out.chomp if $?.success?
+        end
       end
     end
 
--- a/configure
+++ b/configure
@@ -1989,6 +1989,11 @@ Available commands are:
       message.index("\n") != nil
     end
   end
+
+  # Returns true if MacPorts is installed in its default location.
+  def macports?
+    File.exists? '/opt/local/bin/port'
+  end
 end
 
 STDOUT.sync = true
--- a/configure
+++ b/configure
@@ -2000,18 +2000,17 @@ Available commands are:
   end
 
   # Query MacPorts for the path to the latest installed version of
-  # llvm-config.
+  # llvm-config that is within the range of supported LLVM versions.
   def macports_llvm_config
-    # Select latest installed port, as MacPorts can have multiple
-    # versions installed.
-    llvm_port   =
-      %x(port installed | egrep -o 'llvm-[^ ]+' | tail -1).chomp
-    # Query MacPorts for an array of all llvm-config* binaries.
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -1989,6 +1993,25 @@ Available commands are:
       message.index("\n") != nil
     end
   end
+
+  # Returns true if MacPorts is installed in its default location.
+  def macports?
+    File.exists? '/opt/local/bin/port'
```
jc00ke: Why not test for `which port` in case it's installed in a non-default location?

configure
```diff
@@ -1989,6 +1993,25 @@ Available commands are:
...
+  end
+
+  # Query MacPorts for the path to the latest installed version of
+  # llvm-config that is within the range of supported LLVM versions.
+  def macports_llvm_config
+    supported_versions = (3.0 .. 3.5)
+    installed_ports    = %x(port installed | egrep -o 'llvm-[^ ]+' ).split
```
jc00ke: Style-wise, looks like `backticks` are used elsewhere, instead of `%x()`

configure
CodeGnome: @jc00ke The reason is that MacPorts is intended to be installed under /opt/local, while it's possible that there could be some other command named "port" in the PATH. Looking for the canonical location of the MacPorts command seems safe enough; simply assuming that a successful `which port` means MacPorts is actually installed might lead to odd errors later.

configure
CodeGnome: @jc00ke I have no objection to converting the style to backticks. I'll update the pull request later today.

configure
YorickPeterse: While the canonical location might be `/opt/local/bin/port` we have no control over it, and we shouldn't enforce the user to _always_ place MacPorts in a fixed place. One option would be to default to the output of `which port` but allow users to specify a custom path (using something like `--macports=...`.

YorickPeterse: Also, while the argument of duplicate `port` commands in PATH makes sense, this wouldn't work anyway due to the shell commands using a non-absolute path for this command.

CodeGnome: @YorickPeterse That's fine; I'll switch it to just checking the PATH for the *port* command. The current code for Homebrew doesn't even do that; it just calls *brew* and hopes for the best.

I'll post a patch shortly to make this change.

PR #3298, opened 2015-01-28
Search PATH for "port" command.

CodeGnome:
- Make the test for the MacPorts *port* command more flexible by searching the PATH rather than assuming it resides at /opt/local/bin/port. This feature was requested in the line notes for https://github.com/rubinius/rubinius/pull/3297.
- Explicitly test that the *port* command is really MacPorts. This improves reliability beyond the previous reliance on a canonical filesystem location.
- Contrary to expectations, `port version` will return a version string that fails to identify itself as MacPorts. String-matching the interactive prompt is therefore more reliable for validating the executable.

### DIFF
--- a/configure
+++ b/configure
@@ -1994,9 +1994,9 @@ Available commands are:
     end
   end
 
-  # Returns true if MacPorts is installed in its default location.
+  # Returns true if the MacPorts *port* command is found in the PATH.
   def macports?
-    File.exists? '/opt/local/bin/port'
+    system 'which port > /dev/null'
   end
 
   # Query MacPorts for the path to the latest installed version of
--- a/configure
+++ b/configure
@@ -1995,9 +1995,9 @@ Available commands are:
   end
 
   # Returns true if the *port* command is in the PATH and identifies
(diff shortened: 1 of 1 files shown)


jc00ke:
I'm not familiar with MacPorts, but is there a reason we need to match the version number?

CodeGnome:
@jc00ke We don't actually care about the version number except as a textual anchor. It just seems more reliable to match an anchored "MacPorts \<semver>" pattern than just `/MacPorts/`. The goal is defensive coding, not version extraction.

If we care about the MacPorts version in the future, `%x(port version).split.last` is certainly an option, too. However, Rubinius really only cares about MacPorts to the extent that we want to query it for filesystem locations. Unless MacPorts introduces breaking changes to its query interface that we need to branch on, it seems unlikely that we would ever care about the MacPorts version information during the rbx configuration process.

jc00ke:
I didn't think so. Might be more clear just to match `/MacPorts/` or, like you have, `/^MacPorts/`. If the version info isn't used, then we may as well use a simpler regex.

PR #3311, opened 2015-02-08
Add the try_find_const_fast instruction

jemc:
Add the try_find_const_fast instruction,
which returns undefined instead of raising NameError when the constant is not found.

This can be helpful for other languages on the Rubinius VM, trying to implement different constant lookup semantics, but still want to benefit from the `ConstantCache`s.

### DIFF
--- a/vm/llvm/jit_visit.hpp
+++ b/vm/llvm/jit_visit.hpp
@@ -3679,6 +3679,133 @@ use_send:
       stack_push(phi, type::KnownType::constant_cache(constant_cache));
     }
 
+    void visit_try_find_const_fast(opcode& name) {
+      BasicBlock* cont = 0;
+
+      Value* cached_value = 0;
+      BasicBlock* cached_block = 0;
+
+      Value* under = stack_pop();
+
+      ConstantCache** constant_cache_ptr = reinterpret_cast<ConstantCache**>(&name);
+      ConstantCache* constant_cache = *constant_cache_ptr;
+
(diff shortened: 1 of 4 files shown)


vm/llvm/jit_visit.hpp
```diff
@@ -3679,6 +3679,133 @@ namespace rubinius {
...
+                                 offset::CallFrame::constant_scope, "scope_pos"), "frame_scope");
+
+      Value* scope_cmp = b().CreateICmpEQ(constant_scope, frame_scope, "same_scope");
+
+      BasicBlock* check_serial = new_block("check_serial");
+      BasicBlock* check_under  = new_block("check_under");
+      BasicBlock* use_cache    = new_block("use_cache");
+      BasicBlock* use_call     = new_block("use_call");
+
+      cont =      new_block("continue");
```
YorickPeterse: Extra whitespace here shouldn't be needed.

jemc: It was copied verbatim from the find_const_fast visit above.

jemc:
I guess this will fail on Travis until https://github.com/rubinius/rubinius-compiler/pull/3 is added and released.

jemc:
This is not needed for me with Myco anymore, although I think that doing this for Titanius is a good long term goal.  Returning undefined instead of raising an exception for missing constants is an easier-to-use operation.  Something similar for instance variables would be nice too.  Returning undefined instead of nil for missing instance variables is easier to use as well.

I'm closing for now, but we can discuss this more as Titanius begins to take shape.

PR #3315, opened 2015-02-12
Trying to fix #3306 each_slice

### LINKED ISSUE #3306: Mismatch on Enumerator#size result between Rubinius and MRI
Hi all,
Bumped into this situation yesterday.
When calling size on the enumerator returned by calling each_slice with no block, Rubinius and MRI return different values.

MRI 2.1.5, 2.2.0:
```
irb(main):001:0> [1,2,3,4].each_slice(2).size
=> 2
```
RBX 2.5.2
```
irb(main):001:0> [1,2,3,4].each_slice(2).size
=> nil
```

As a side note, [jruby 9000 pre1 also returns nil](https://github.com/jruby/jruby/issues/2553).

rbx version:
```
rubinius 2.5.2 (2.1.0 7a5b05b1 2015-01-30 3.5.1 JI) [x86_64-darwin14.1.0]
```

OS:
```
14.1.0 Darwin Kernel Version 14.1.0: Mon Dec 22 23:10:38 PST 2014; root:xnu-2782.10.72~2/RELEASE_X86_64 x86_64
```

Thank you

bjfish:
Tried to understand enum.c to fix #3306 and produce something similar where a size
function can be specified.

The size functions there are a little difficult to follow and the enum library here might need more refactoring in this area. This is just a taking a shot at fixing the issue without causing regression.

Forgive me for not reading all the contributing guidelines yet but I did run all the specs (I think).

### DIFF
--- a/kernel/common/enumerator.rb
+++ b/kernel/common/enumerator.rb
@@ -151,7 +151,7 @@ module Enumerable
     def size
       if @size.kind_of?(Proc)
         @size.call
-      elsif @size.is_a? Numeric
+      elsif @size.kind_of?(Numeric)
         @size
       elsif @size.respond_to? :size
         @size.size
--- a/kernel/common/enumerator.rb
+++ b/kernel/common/enumerator.rb
@@ -35,6 +35,7 @@ module Enumerable
           raise ArgumentError, "Enumerator#initialize requires a block when called without arguments"
         end
 
+        size = receiver_or_size
         receiver = receiver_or_size
       end
 
@@ -148,7 +149,19 @@ module Enumerable
     end
 
     def size
-      @size.kind_of?(Proc) ? @size.call : @size
+      if @size.kind_of?(Proc)
+        @size.call
+      elsif @size.is_a? Numeric
+        @size
+      elsif @size.respond_to? :size
+        @size.size
+      else
+        @size
+      end
+    end
+
+    def size=(size)
+      @size = size
     end
 
     def with_index(offset=0)
--- a/kernel/common/enumerator.rb
+++ b/kernel/common/enumerator.rb
@@ -149,15 +149,7 @@ module Enumerable
     end
 
     def size
-      if @size.kind_of?(Proc)
-        @size.call
-      elsif @size.kind_of?(Numeric)
-        @size
-      elsif @size.respond_to? :size
-        @size.size
-      else
-        @size
-      end
+      @size.kind_of?(Proc) ? @size.call : @size
     end
 
     def size=(size)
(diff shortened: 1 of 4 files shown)


kernel/common/enumerator.rb
```diff
@@ -148,7 +149,19 @@ def rewind
...
 
     def size
-      @size.kind_of?(Proc) ? @size.call : @size
+      if @size.kind_of?(Proc)
+        @size.call
+      elsif @size.is_a? Numeric
+        @size
+      elsif @size.respond_to? :size
+        @size.size
+      else
```
YorickPeterse: What is the reason for changing this method itself? As far as I'm aware `@size` is always either a Numeric or a Proc, thus it will never respond to `size`. Checking if it's a `Numeric` also isn't really needed since we can guarantee the types of `@size`.

kernel/common/enumerator.rb
```diff
@@ -148,7 +149,19 @@ def rewind
     end
 
     def size
-      @size.kind_of?(Proc) ? @size.call : @size
+      if @size.kind_of?(Proc)
+        @size.call
+      elsif @size.is_a? Numeric
```
jemc: As a minor style note, I'd suggest using `kind_of?(Numeric)` here be consistent with `kind_of?(Proc)`

YorickPeterse:
The core problem of this bug is that `@size` is set to `nil` when using `Enumerator#each_slice`. Because of this the `@size.kind_of?(Proc)` check evaluates to false, leading to the `#size` method just returning `@size` (and thus `nil`).

From the top of my head, an easy way to probably fix this would be to simply use the following:

```ruby
def size
  if @size.kind_of?(Proc)
    @size.call
  elsif @size
    @size
  else
    count
  end
end
```

Since `Enumerator#count` _does_ work properly on an Enumerator returned by `each_slice` this will most likely do the trick.

spec/ruby/core/enumerable/each_slice_spec.rb
```diff
@@ -56,4 +56,10 @@
     multi = EnumerableSpecs::YieldsMulti.new
     multi.each_slice(2).to_a.should == [[[1, 2], [3, 4, 5]], [[6, 7, 8, 9]]]
   end
+
+  it "has correct size with no proc given" do
```
YorickPeterse: Spec descriptions should read as natural English. While quite a few existing descriptions don't follow this guideline we try to apply it to new changes. As such I'd suggest writing this something along the lines of `it "returns the correct size when no block is given"`.

kernel/common/enumerator.rb
bjfish: @YorickPeterse in the case of each_with_index it was returning the array, there may be another (more correct) way to resolve this
```
[1, 2, 3, 5].each_with_index.size
-> [1, 2, 3, 5]
```

YorickPeterse: You mean `size` was returning an Array? If so, that should probably belong in its own PR (unless it's directly related to `each_slice`).

PR #3318, opened 2015-02-15
Spec size method calls on Enumerator instances returned by Enumerable methods.

fmfdias:
Hi,

A few days ago I've opened [this issue](https://github.com/rubinius/rubinius/issues/3306) regarding calling size on the Enumerator instance that results on calling the Enumerable#each_slice without a block.

At the time I only checked that method. A few days after I thought on checking other similar Enumerable methods that return an Enumerator, on MRI 2.1.5, Rubinius 2.5.2 and JRuby 9.0.0.0 pre1.
Rubinius returns nil on all the calls which is not the behavior of MRI. JRuby was only failing on each_slide.
Unfortunately I wasn't able to finish it and inform you guys of it before the issue was handled and closed.

I took a look at the tests that were added for the each_slice fix [pull request](https://github.com/rubinius/rubinius/pull/3315) and, based on it, decided to transform the checks I made in specs that could be useful.

I tried to cover all the Enumerable methods that I could find. I also checked the Array class and added similar tests to the methods implemented/overridden by the Array class that are on the same situation.

I hope it helps.

Thank you!

### DIFF
--- a/spec/ruby/core/enumerable/drop_while_spec.rb
+++ b/spec/ruby/core/enumerable/drop_while_spec.rb
@@ -45,7 +45,7 @@ describe "Enumerable#drop_while" do
     multi.drop_while {|e| e != [6, 7, 8, 9] }.should == [[6, 7, 8, 9]]
   end
 
-  it "returns the nil as size when no block is given" do
+  it "returns nil as size when no block is given" do
     enum = EnumerableSpecs::NumerousWithSize.new(1, 2, 3, 4, 5, 0)
     enum.drop_while.size.should == nil
   end
--- a/spec/ruby/core/enumerable/drop_while_spec.rb
+++ b/spec/ruby/core/enumerable/drop_while_spec.rb
@@ -44,4 +44,9 @@ describe "Enumerable#drop_while" do
     multi = EnumerableSpecs::YieldsMulti.new
     multi.drop_while {|e| e != [6, 7, 8, 9] }.should == [[6, 7, 8, 9]]
   end
+
+  it "returns the nil as size when no block is given" do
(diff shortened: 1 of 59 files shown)


jemc:
@fmfdias 
Thanks for taking the time to write some specs!  However, we need to tag these as failing so that they don't cause CI or`rake spec` failures.

Take a look at `mspec tag --help` for more information, but the most basic tagging of failures looks something like `mspec tag fails ./spec`.  This will add files or lines to the directory structure in `./spec/tags` that mirrors the structure of the specs in `./spec`.

fmfdias:
Hi @jemc 

I've tagged them as failing.
Thank you for the feedback.

Cheers

YorickPeterse:
Travis seems to fail due to a race condition unrelated to these changes (https://travis-ci.org/rubinius/rubinius/jobs/50984264), looks good to me :+1:. Thanks!

spec/ruby/core/enumerable/drop_while_spec.rb
```diff
@@ -44,4 +44,9 @@
     multi = EnumerableSpecs::YieldsMulti.new
     multi.drop_while {|e| e != [6, 7, 8, 9] }.should == [[6, 7, 8, 9]]
   end
+
+  it "returns the nil as size when no block is given" do
```
YorickPeterse: I think this should just be "returns nil as size...", unless `nil` is suddenly a person/object :smile:

fmfdias: Got caught, hehehe. I think it could also give a nice a movie title... 'The Nil' :smile:

Issue #3269, opened 2015-01-04
Process.spawn specs raise Errno::EBADF instead of Errno::EACCESS


YorickPeterse:
On Arch Linux the `Process.spawn` specs fail due to raising `EBADF` instead of `EACCESS`. This is using LLVM 3.5 although I'd be surprised if LLVM versions are even related to this problem.

Test output:

```
Running 651 tests.................................................................................................................................................................................................................
..................................................................................................................................................................................................................................
........................................................................................................................................................................................................................OK!
bin/mspec ci :ci_files  -t bin/rbx -d --background
rubinius 2.4.1.n349 (2.1.0 fbaafa65 2014-12-15 3.5.0 JI) [x86_64-linux-gnu]
..................................................................................................................................................................................................................................
..................................................................................................................................................................................................................................
..................................................................................................................................................................................................................................
..................................................................................................................................................................................................................................
......F...................................................................................F.......................................................................................................................................
............................................................................................................................................................................F.....................................................
..................................................................................................................................................................................................................................
..................................................................................................................................................................................................................................
..................................................................................................................................................................................................................................
................................................................................................................................

1)
Kernel#` raises an Errno::ENOENT if the command is not executable FAILED
Expected Errno::ENOENT
but got Errno::EBADF (Bad file descriptor - execvp(2) failed)
          SpecExpectation.fail_with at mspec/lib/mspec/expectations/expectations.rb:15
                Object(Proc)#should at mspec/lib/mspec/expectations/should.rb:8
           { } in Object#__script__ at spec/ruby/core/kernel/backtick_spec.rb:28
  BasicObject(Object)#instance_eval at kernel/common/eval.rb:43
                      MSpec.protect at mspec/lib/mspec/runner/mspec.rb:69
        { } in ContextState#protect at mspec/lib/mspec/runner/context.rb:179
      { } in Enumerable(Array)#all? at kernel/common/enumerable.rb:329
                         Array#each at kernel/bootstrap/array.rb:76
             Enumerable(Array)#all? at kernel/common/enumerable.rb:329
               ContextState#protect at mspec/lib/mspec/runner/context.rb:179
        { } in ContextState#process at mspec/lib/mspec/runner/context.rb:208
                { } in MSpec.repeat at mspec/lib/mspec/runner/mspec.rb:239
              Integer(Fixnum)#times at kernel/common/integer.rb:196
                       MSpec.repeat at mspec/lib/mspec/runner/mspec.rb:238
[...]

Issue #3264, opened 2015-01-01
Need Ruby 2.2 specs


YorickPeterse:
Pulled from the [NEWS](https://github.com/ruby/ruby/blob/v2_2_0_preview1/NEWS) file

* nil/true/false
  - [x] nil/true/false objects are frozen - spec: 0d0d809c292b7e1a2ed66479f0b227a5fce17f9e & implementation: 6580956db009c0e376ead78067639d40d56adaf7

* Binding
  * New methods:
    - [x] Binding#local_variables - spec: 81b9f5d6bbfa6b8df8ab722098981f2856b07a7d & implementation: 32f555ef9a198e5a519fd5c1cf70024cb05d5eb1
    - [x] Binding#receiver - spec: 0c9027b8d4705481d641422cdcc13e031a91f7cf & implementation: 32f555ef9a198e5a519fd5c1cf70024cb05d5eb1

* Dir
  * New methods:
    - [x] Dir#fileno - spec: 96d033d303b0ea6aa3b661523556f4bb87156829, 2958f37ffe8cf620bbd23e3a0c1b1f1872592b43 & implementation: 7111811f6456daea686de6f0adafa3dacdd3f3ed

* Enumerable
  * New methods:
    - [x] Enumerable#slice_after - spec: 73496c350473a05b721745bea0cd8ca4da9fb71b & implementation: 6590ec7b285ad1a87651ee17f096f95de2b7eb94
    - [x] Enumerable#slice_when - spec: 9ebed9ab6d5ba704977342d0f77ea933550868b3 & implementation: f21d58d735383a30156575a2c51b0982b6bae217
  * Extended methods:
    - [x] min (supports optional argument to return multiple elements) - spec: 47f8941210f39daf0bb4f6ecee8a05dde689afbd & implementation: #3420
    - [x] min_by (supports optional argument to return multiple elements) - spec: 56cd372e68d1477d6938a6162a6521a983fa0eb6, 17f2186d47f9f1ad9bba36aee232deeb60a7d2c2 & implementation: #3420
    - [x] max (supports optional argument to return multiple elements) - spec: ef33bafa58594989851ca864cb63afa51e540700 & implementation: #3420
    - [x] max_by (supports optional argument to return multiple elements) - spec: a8d91ea46ebf667477bcab849efb7dc638b52616, 17f2186d47f9f1ad9bba36aee232deeb60a7d2c2 & implementation: #3420

* Float
  * New methods:
[...]

PR #3320, opened 2015-02-18
Fix enumerator size

fmfdias:
Hi,

This is an attempt to fix the Enumerator#size method issues identified previously [here](https://github.com/rubinius/rubinius/issues/3306) and [here](https://github.com/rubinius/rubinius/pull/3318). It's still a work in progress but I opened this pull request to check with you guys if I'm on the right track and also get some guidance.

I took a look at the JRuby implementation and what they do is pass the a 'calculated' size to the enumerator. Usually the size is enough (each, reverse_each, etc...), but there are some methods that need some math, like the cycle, each_cons and each_slice.

To start I've implemented a to_enum_with_size private method on the Enumerable module to use on these methods to create the an enum with the size calculation. Doing it here allow it to be used on Array and other classes that mix in the Enumerable module. Nevertheless, my question here is if it's the right place to do it.

Another thing I would like to point out is the each_cons and each_slice size spec. The size calculation is a little bit more complex that the cycle. The spec is only testing some value combinations, but I think it could be better. Do you guys have a better way to implement it?

When I was at it, I saw that both each_cons and each_slice raise an exception when called without a block. I added the calls to the spec that tests that the exception is raised and fixed it.

Like I said, this is still a work in progress. Enumerable seems to be fixed with these changes, but Array and other aren't. I already started with Array. With time, and if you are all ok with these changes, I'll continue to add them to this pull request.

Thank you

### DIFF
--- a/kernel/common/array.rb
+++ b/kernel/common/array.rb
@@ -392,7 +392,12 @@ class Array
 
   def combination(num)
     num = Rubinius::Type.coerce_to_collection_index num
-    return to_enum(:combination, num) unless block_given?
+
+    unless block_given?
+      return to_enum(:combination, num) do
+        Rubinius::Mirror::Array.reflect(self).combination_size(num)
+      end
+    end
 
     if num == 0
       yield []
@@ -470,7 +475,12 @@ class Array
   end
 
   def cycle(n=nil)
-    return to_enum(:cycle, n) unless block_given?
+    unless block_given?
+      return to_enum(:cycle, n) do
+        Rubinius::EnumerableHelper.cycle_size(size, n)
+      end
+    end
+
     return nil if empty?
 
     # Don't use nil? because, historically, lame code has overridden that method
@@ -546,7 +556,7 @@ class Array
   end
 
   def delete_if
-    return to_enum(:delete_if) unless block_given?
+    return to_enum(:delete_if) { size } unless block_given?
 
     Rubinius.check_frozen
 
@@ -574,7 +584,7 @@ class Array
   end
 
   def each_index
-    return to_enum(:each_index) unless block_given?
+    return to_enum(:each_index) { size } unless block_given?
 
     i = 0
     total = @total
@@ -938,7 +948,7 @@ class Array
   end
 
   def keep_if(&block)
-    return to_enum :keep_if unless block_given?
+    return to_enum(:keep_if) { size } unless block_given?
 
     Rubinius.check_frozen
 
@@ -982,7 +992,11 @@ class Array
   end
 
   def permutation(num=undefined, &block)
-    return to_enum(:permutation, num) unless block_given?
+    unless block_given?
+      return to_enum(:permutation, num) do 
+        Rubinius::Mirror::Array.reflect(self).permutation_size(num)
+      end
+    end
 
     if undefined.equal? num
       num = @total
@@ -1134,12 +1148,12 @@ class Array
   end
 
   def reject(&block)
-    return to_enum(:reject) unless block_given?
+    return to_enum(:reject) { size } unless block_given?
     Array.new(self).delete_if(&block)
   end
 
   def reject!(&block)
-    return to_enum(:reject!) unless block_given?
+    return to_enum(:reject!) { size } unless block_given?
 
     Rubinius.check_frozen
 
@@ -1153,7 +1167,9 @@ class Array
   def repeated_combination(combination_size, &block)
     combination_size = combination_size.to_i
     unless block_given?
-      return Enumerator.new(self, :repeated_combination, combination_size)
+      return to_enum(:repeated_combination, combination_size) do
+        Rubinius::Mirror::Array.reflect(self).repeated_combination_size(combination_size)
+      end
     end
 
     if combination_size < 0
@@ -1183,7 +1199,9 @@ class Array
   def repeated_permutation(combination_size, &block)
     combination_size = combination_size.to_i
     unless block_given?
-      return Enumerator.new(self, :repeated_permutation, combination_size)
+      return to_enum(:repeated_permutation, combination_size) do
+        Rubinius::Mirror::Array.reflect(self).repeated_permutation_size(combination_size)
+      end
     end
 
     if combination_size < 0
@@ -1227,7 +1245,7 @@ class Array
   end
 
   def reverse_each
-    return to_enum(:reverse_each) unless block_given?
+    return to_enum(:reverse_each) { size } unless block_given?
 
     stop = @start - 1
     i = stop + @total
@@ -1334,7 +1352,7 @@ class Array
   end
 
   def select!(&block)
-    return to_enum :select! unless block_given?
+    return to_enum(:select!) { size } unless block_given?
 
     Rubinius.check_frozen
 
@@ -1620,7 +1638,7 @@ class Array
   def sort_by!(&block)
     Rubinius.check_frozen
 
-    return to_enum :sort_by! unless block_given?
+    return to_enum(:sort_by!) { size } unless block_given?
 
     replace sort_by(&block)
   end
--- a/kernel/common/enumerable.rb
+++ b/kernel/common/enumerable.rb
@@ -50,7 +50,7 @@ module Enumerable
       end
       ary
     else
-      to_enum :collect
+      to_enum(:collect) { enumerator_size }
     end
   end
 
@@ -69,7 +69,7 @@ module Enumerable
   end
 
   def each_entry(*pass)
-    return to_enum :each_entry, *pass unless block_given?
+    return to_enum(:each_entry, *pass) { enumerator_size } unless block_given?
     each(*pass) do |*args|
       yield args.size == 1 ? args[0] : args
     end
@@ -77,7 +77,7 @@ module Enumerable
   end
 
   def each_with_object(memo)
-    return to_enum :each_with_object, memo unless block_given?
+    return to_enum(:each_with_object, memo) { enumerator_size } unless block_given?
     each do
       obj = Rubinius.single_block_arg
       yield obj, memo
@@ -88,7 +88,7 @@ module Enumerable
   alias_method :with_object, :each_with_object
 
   def flat_map
-    return to_enum(:flat_map) unless block_given?
+    return to_enum(:flat_map) { enumerator_size } unless block_given?
 
     array = []
     each do |*args|
@@ -120,7 +120,7 @@ module Enumerable
   private :enumerator_size
 
   def group_by
-    return to_enum(:group_by) unless block_given?
+    return to_enum(:group_by) { enumerator_size } unless block_given?
 
     h = {}
     each do
@@ -226,7 +226,7 @@ module Enumerable
   end
 
   def each_with_index(*args)
-    return to_enum(:each_with_index, *args) unless block_given?
+    return to_enum(:each_with_index, *args) { enumerator_size } unless block_given?
 
     idx = 0
     each(*args) do
@@ -281,7 +281,7 @@ module Enumerable
   end
 
   def sort_by
-    return to_enum :sort_by unless block_given?
+    return to_enum(:sort_by) { enumerator_size } unless block_given?
 
     # Transform each value to a tuple with the value and it's sort by value
     sort_values = map do
@@ -351,7 +351,11 @@ module Enumerable
   end
 
   def cycle(many=nil)
-    return to_enum(:cycle, many) unless block_given?
+    unless block_given?
+      return to_enum(:cycle, many) do
+        Rubinius::EnumerableHelper.cycle_size(enumerator_size, many)
+      end
+    end
 
     if many
       many = Rubinius::Type.coerce_to_collection_index many
@@ -408,10 +412,21 @@ module Enumerable
   end
 
   def each_cons(num)
-    return to_enum(:each_cons, num) unless block_given?
-
     n = Rubinius::Type.coerce_to_collection_index num
     raise ArgumentError, "invalid size: #{n}" if n <= 0
+
+    unless block_given?
+      return to_enum(:each_cons, num) do
+        enum_size = enumerator_size
+        if enum_size.nil? 
+          nil 
+        elsif enum_size == 0
+          0
+        else
+          enum_size - n + 1
+        end
+      end
+    end
 
     array = []
     each do
@@ -424,20 +439,15 @@ module Enumerable
   end
 
   def each_slice(slice_size)
-    unless block_given?
-      enum = to_enum(:each_slice, slice_size)
-
-      # For non lazy enumerators the enum size must be set to the size of the
-      # slice, _not_ the size of the containing enumerator.
-      unless enum.kind_of?(Enumerator::Lazy)
-        Rubinius.privately { enum.size = enum.count }
-      end
-
-      return enum
-    end
-
     n = Rubinius::Type.coerce_to_collection_index slice_size
     raise ArgumentError, "invalid slice size: #{n}" if n <= 0
+
+    unless block_given?
+      return to_enum(:each_slice, slice_size) do
+        enum_size = enumerator_size
+        enum_size.nil? ? nil : (enum_size.to_f / n).ceil
+      end
+    end
 
     a = []
     each do
@@ -467,7 +477,7 @@ module Enumerable
   alias_method :detect, :find
 
   def find_all
-    return to_enum(:find_all) unless block_given?
+    return to_enum(:find_all) { enumerator_size } unless block_given?
 
     ary = []
     each do
@@ -548,7 +558,7 @@ module Enumerable
   end
 
   def max_by
-    return to_enum(:max_by) unless block_given?
+    return to_enum(:max_by) { enumerator_size } unless block_given?
 
     max_object = nil
     max_result = undefined
@@ -568,7 +578,7 @@ module Enumerable
   end
 
   def min_by
-    return to_enum(:min_by) unless block_given?
+    return to_enum(:min_by) { enumerator_size } unless block_given?
 
     min_object = nil
     min_result = undefined
@@ -623,7 +633,7 @@ module Enumerable
   end
 
   def minmax_by(&block)
-    return to_enum(:minmax_by) unless block_given?
+    return to_enum(:minmax_by) { enumerator_size } unless block_given?
 
     min_object = nil
     min_result = undefined
@@ -684,7 +694,7 @@ module Enumerable
   end
 
   def partition
-    return to_enum(:partition) unless block_given?
+    return to_enum(:partition) { enumerator_size } unless block_given?
 
     left = []
     right = []
@@ -697,7 +707,7 @@ module Enumerable
   end
 
   def reject
-    return to_enum(:reject) unless block_given?
+    return to_enum(:reject) { enumerator_size } unless block_given?
 
     ary = []
     each do
@@ -709,7 +719,7 @@ module Enumerable
   end
 
   def reverse_each(&block)
-    return to_enum(:reverse_each) unless block_given?
+    return to_enum(:reverse_each) { enumerator_size } unless block_given?
 
     # There is no other way then to convert to an array first... see 1.9's source.
     to_a.reverse_each(&block)
--- a/kernel/common/kernel.rb
+++ b/kernel/common/kernel.rb
@@ -504,7 +504,7 @@ module Kernel
   module_function :load
 
   def loop
-    return to_enum(:loop) unless block_given?
+    return to_enum(:loop) { Float::INFINITY } unless block_given?
 
     begin
       while true
--- a/spec/ruby/core/array/permutation_spec.rb
+++ b/spec/ruby/core/array/permutation_spec.rb
@@ -106,4 +106,33 @@ describe "Array#permutation" do
 
     accum.should == [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 1, 2], [3, 2, 1]]
   end
+
+  describe "when no block is given" do
+    describe "returned Enumerator" do
+      describe "size" do
+        describe "with an array size greater than 0" do
+          it "returns the descending factorial of array size and given length" do
+            @numbers.permutation(4).size.should == 0
+            @numbers.permutation(3).size.should == 6
+            @numbers.permutation(2).size.should == 6
+            @numbers.permutation(1).size.should == 3
+            @numbers.permutation(0).size.should == 1
+          end
+          it "returns the descending factorial of array size with array size when there's no param" do
+            @numbers.permutation.size.should == 6
+            [1,2,3,4].permutation.size.should == 24
+            [1].permutation.size.should == 1
+          end
+        end
+        describe "with an empty array" do
+          it "returns 1 when the given length is 0" do
+            [].permutation(0).size.should == 1
+          end
+          it "returns 1 when there's param" do
+            [].permutation.size.should == 1
+          end
+        end
+      end
+    end
+  end
 end
--- a/kernel/bootstrap/array.rb
+++ b/kernel/bootstrap/array.rb
@@ -66,7 +66,7 @@ class Array
   # Passes each element in the Array to the given block
   # and returns self.
   def each
-    return to_enum(:each) unless block_given?
+    return to_enum(:each) { size } unless block_given?
 
     i = @start
     total = i + @total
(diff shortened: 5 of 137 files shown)


kernel/common/enumerable.rb
```diff
@@ -351,7 +351,18 @@ def any?
   end
 
   def cycle(many=nil)
-    return to_enum(:cycle, many) unless block_given?
+    unless block_given?
+      size = Proc.new do
```
YorickPeterse: This pattern occurs in a few places in this PR so this comment applies to all of them: can we not simply pass the size itself here instead of a Proc? There's little overhead in calculating the size at this point and it removes the need for evaluating the block on every `#size` call later on.

kernel/common/enumerable.rb
```diff
@@ -424,21 +440,16 @@ def each_cons(num)
...
+    n = Rubinius::Type.coerce_to_collection_index slice_size
+    raise ArgumentError, "invalid slice size: #{n}" if n <= 0
 
-      # For non lazy enumerators the enum size must be set to the size of the
-      # slice, _not_ the size of the containing enumerator.
-      unless enum.kind_of?(Enumerator::Lazy)
-        Rubinius.privately { enum.size = enum.count }
+    unless block_given?
+      size = Proc.new do
+        (enum_size = enumerator_size).nil? ? nil : (enum_size.to_f / n).ceil
```
YorickPeterse: I'm not sure if it's 100% safe here to rely on a slice size always being of size `total_enum_size / slice_size`, granted I can't think of a possibility where this is not the case.

kernel/common/enumerable.rb
```diff
@@ -754,4 +765,13 @@ def include?(obj)
 
   alias_method :member?, :include?
 
+  def to_enum_with_size(method, size, *args)
```
YorickPeterse: I'm not sure if this method belongs in Enumerable itself, we might want to use a mirror for it instead (not sure if they work with modules though). Also, I think it's better if we clean up the mess of `Enumerator#initialize` and `Enumerator#initialize_enumerator` instead of using allocate and the latter. This however is up to @brixen since I'm not 100% sure of it.

YorickPeterse:
Spec changes look fine to me. For Enumerable/Enumerator we might want to do some general book keeping to clean things up, especially regarding passing in enum sizes.

YorickPeterse:
To further iterate (now that I'm at the office): I think we should fix `Enumerator#initialize` so we don't need `#initialize_enumerator` anymore. Right now the former method changes behaviour depending on whether or not you give it a block and uses the same argument as either the receiver or the size. The argument list I'm thinking of as an alternative would be:

* `size`
* `receiver`
* `method_name`
* `*method_args`
* `&block`

This would still match the signature of MRI (which is `Enumerator#initialize(size)`) and should remove the need for the extra `initialize_enumerator` method.

kernel/common/enumerable.rb
fmfdias: I agree with you! These sizes are not that complex to calculate. There will be others, like the [Array#combination](https://github.com/ruby/ruby/blob/v2_1_5/array.c#L4850) that might be worthwhile calculating it only if needed. Will see when we get there, but in the meantime I'll change these.

kernel/common/enumerable.rb
fmfdias: I'll double check, just in case... :smile:

fmfdias:
Hi @YorickPeterse,

Thanks for the input.
I've replied to some of your comments inline. Regarding the need of the to_enum_with_size method I'm open to try other approaches. I also don't feel to much comfortable about it.
In the meantime I'll try the approach described in the [to_enum documentation](http://ruby-doc.org/core-2.1.1/Object.html#method-i-to_enum).

I'm doing this off hours, and with some time constraints. That's why I'm only replying now. My main objective is to help and so because I like this, but if you guys feel that the time might be an issue, please go ahead and continue with it as you see better. I'll continue to help in other ways.

Will get back when I have more news.
Thank you

YorickPeterse:
There's no rush, so take your time :)

fmfdias:
Hi @YorickPeterse,

Thanks!
Reimplemented it with the to_enum and removed the to_enum_with_size methods and it works :)
There's some new edge case checks for the specs of the methods that involve some math in the calculation.
Before going to the Array and others, I'm thinking on improving the added specs.
The reasons are:
* One spec is missing is that the size is always nil when the Enumerable doesn't implement a size method.
* Until now I saw 3 repeating patterns:
  * The resulting enumerator size returns the enumerable size (collect, find_all, etc...)
  * The resulting enumerator size returns nil (find_index, take_while, etc...)
  * The resulting enumerator size returns has some specific math involved (cycle, each_slice, each_cons)
* Array, and probably others, seem to also present similar patterns (e.g: Array#each_index) and they override some Enumerable methods (e.g.: Array#cycle).

Doing this would greatly reduce the amount of work of creating/changing the specs. I was thinking on using it_behaves_like but I don't if it's the best way to do it. There's also the question of where to place in the case of having to share them with Enumerables like Array. I'll try to figure it out, but any advice would be great.

Thank you

YorickPeterse:
For re-using specs we use shared examples, e.g. https://github.com/rubinius/rubinius/blob/master/spec/ruby/shared/enumerator/with_object.rb. This can then be used as following: https://github.com/rubinius/rubinius/blob/master/spec/ruby/core/enumerator/with_object_spec.rb#L5. This is probably the best solution here since Array and Enumerable share the same expected behaviour.

Regarding the patterns, make sure the changes also take care of lazy enumerators as in that case `size` is always supposed to return `nil` if I remember correctly.

bjfish:
I had a failing spec I forgot to add before for each_with_index_spec:
```ruby
 it "has correct size with no proc given" do
    [1, 2, 3, 5].each_with_index.size.should == 4
  end
```
@fmfdias I thought I'd mention it in case you want to add it.

fmfdias:
Hi guys,

Thanks for the feedback.
Sorry for the late reply but had a hard drive failure this week and had less time to look at this.

I've updated the specs with two shared spec files. The enumeratorized is to be use by any enumerable implementation and besides implementing the 2 first use cases I enumerated in my last comment, it also has the cycle case that's shared with the Array. The enumerable_enumeratorized is specific to the Enumerable specs and basically adds tests for Enumerable with no size method implementation.

The usage of this specs relies on it_behaves_like on most cases and in some edge cases it relies on the it_should_behave_like (e.g: Other shared specs, each_slice, each_cons...)

@YorickPeterse Thanks for the ideas. I also checked with MRI 2.1.5 and size is also passed to lazy enumerators. This version also does that :)

@bjfish Thank you, this version also has a spec for each_with_index. If you remember any other that might be missing, please tell me and I'll add.

For next steps I'm going to add some missing specs on Array (e.g: combination, permutation, etc...) and start implementing them.

Thanks

fmfdias:
Hi,

Added the implementation for the Array methods that were failing the specs.

I moved to Hash... added failing specs and started to implement it on hash.rb when I found out that there's another hash implementation (hash_hamt.rb).
This might seem like a noob question, but... how do enable it / test it? :)

Thank you

jemc:
@fmfdias 
Try `RBXOPT=-Xhash.hamt`

fmfdias:
@jemc Thank you!

The Hash part is implemented. Moving to Range!

Cheers

fmfdias:
Hi guys,

A little status update...
Range, Kernel and Numeric methods with to_enum calls are fixed and have tests. At this moment I'm checking all the to_enum calls and comparing them to the MRI implementation, so it might take longer than I expected to finish this pull request.
This weekend I hope to advance a bit on this!

Have a nice weekend!

YorickPeterse:
Thanks for looking into this, and take your time! :+1:

fmfdias:
Hi guys,

@YorickPeterse: Thanks :)

I think the bulk work is done. I tried to find, check and fix every Kernel#to_enum, Kernel#enum_for and Enumerator#new existing on the Rubinius code. Hope none has skipped, hehehe.

Please review and I'll make the necessary changes as soon as I can.

Have a nice weekend!

YorickPeterse:
I haven't looked at the actual code yet (I'll do so in the next few days), but looking at the commits themselves I have two nit-picky comments:

1. The merge commit https://github.com/fmfdias/rubinius/commit/e809904294c86099c13338d17873547685cd88a4 can be removed using a git rebase if I'm not mistaken. It's not hugely crucial but it keeps the logs clean.
2. Whenever possible we prefer for the first line of a Git commit to be <= 50 characters. Given such a summary is clear enough it's much easier to quickly scan through commits to see what they're up to.

Both these changes can be made using an interactive Git rebase (`git rebase -i HEAD~53` where `53` is the number of latest commits to rebase). If you're new to rebasing I suggest reading http://git-scm.com/book/en/v2/Git-Branching-Rebasing since it's pretty useful (although a bit hard to grok at first). If it's too much of a fuzz I can take a look at it and explain how/what I did :)

kernel/common/array.rb
```diff
@@ -390,9 +390,25 @@ def clear
...
   end
 
+  # Implementation based on the binomial_coefficient
+  # implemented on MRI array.c.
+  def __binomial_coefficient__(comb, size)
+    comb = size-comb if (comb > size-comb)
+    return 0 if comb < 0
+    __descending_factorial__(size, comb) / __descending_factorial__(comb, comb)
+  end
+  private :__binomial_coefficient__
```
YorickPeterse: This should probably be moved into a mirror for the Array class (https://github.com/rubinius/rubinius/blob/master/kernel/bootstrap/array_mirror.rb), that way we don't pollute the Array class nor can Ruby code in the wild mess with it. Mirrors can be used as following: https://github.com/rubinius/rubinius/blob/master/kernel/common/array.rb#L48

kernel/common/array.rb
```diff
@@ -981,8 +997,37 @@ def pack(directives)
...
+  def __descending_factorial__(from, how_many)
+    cnt = how_many >= 0 ? 1 : 0
+    while (how_many) > 0
+      cnt = cnt*(from)
+      from -= 1
+      how_many -= 1
+    end
+    cnt
+  end
+  private :__descending_factorial__
```
YorickPeterse: Same story here as with `__binomial_coefficient__`.

kernel/common/array.rb
```diff
@@ -1150,10 +1195,16 @@ def reject!(&block)
     self
   end
 
+  def __repeated_combination_size__(combination_size)
```
YorickPeterse: Same mirror story here too.

kernel/common/array.rb
```diff
@@ -1180,10 +1231,16 @@ def compile_repeated_combinations(combination_size, place, index, depth, &block)
 
   private :compile_repeated_combinations
 
+  def __repeated_permutation_size__(combination_size)
```
YorickPeterse: And same here as well.

kernel/common/enumerable.rb
```diff
@@ -350,8 +350,19 @@ def any?
     false
   end
 
+  def __cycle_size__(enum_size, many)
```
YorickPeterse: Same here regarding mirrors.

kernel/common/enumerable.rb
```diff
@@ -408,11 +419,19 @@ def drop_while
   end
 
   def each_cons(num)
-    return to_enum(:each_cons, num) unless block_given?
-
     n = Rubinius::Type.coerce_to_collection_index num
     raise ArgumentError, "invalid size: #{n}" if n <= 0
 
+    return to_enum(:each_cons, num) {
+      if (enum_size = enumerator_size).nil?
```
YorickPeterse: What is the reason for the assignment here, can't this just use `if enumerator_size.nil?`?

YorickPeterse: Ah wait, I think this means that if it _is_ a non-nil value it can be used i the other branches in the if statement. In that case I think it makes more sense to just use `enumerator_size` directly as using an assignment in an if statement is a bit confusing.

kernel/common/enumerable.rb
```diff
@@ -424,21 +443,13 @@ def each_cons(num)
...
-      end
-
-      return enum
-    end
-
     n = Rubinius::Type.coerce_to_collection_index slice_size
     raise ArgumentError, "invalid slice size: #{n}" if n <= 0
 
+    return to_enum(:each_slice, slice_size) {
+      (enum_size = enumerator_size).nil? ? nil : (enum_size.to_f / n).ceil
```
YorickPeterse: Same here, unless I'm mistaken about the purpose of the assignment this code can just use `enumerator_size` directly.

kernel/common/kernel.rb
```diff
@@ -504,7 +504,7 @@ def load(name, wrap=false)
   module_function :load
 
   def loop
-    return to_enum(:loop) unless block_given?
+    return to_enum(:loop) { Float::INFINITY } unless block_given?
```
YorickPeterse: Interesting difference: `to_enum(:loop).size` returns `nil` on MRI 2.2, but using `Kernel.loop.size` returns `Infinity`. As far as I can tell this is implementing `Kernel.loop` so this should be fine.

spec/ruby/core/array/permutation_spec.rb
```diff
@@ -106,4 +106,33 @@
 
     accum.should == [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 1, 2], [3, 2, 1]]
   end
+
+  describe "when no block is given" do
+    describe "returned Enumerator" do
+      describe "size" do
+        describe "array size > 0" do
```
YorickPeterse: nit-picky: something like `describe "with an array size greater than 0"` reads a bit more natural.

YorickPeterse:
After looking through the code I have two general remarks:

1. As mentioned in the individual comments it's probably best to remove the private `__XXX__` methods into a mirror class (whenever possible). This ensures third-party Ruby code can't mess with it.
2. How do these changes handle lazy enumerators? Do the existing lazy enumerator specs still pass?

YorickPeterse:
Some more info/rationale on mirrors can be found in commit 06d6f25c5fa18ce667b80ba2aa77449b24b72c23.

YorickPeterse:
@brixen it seems we currently don't support mirrors for modules (e.g. `Enumerable`) as `Rubinius::Mirror.reflect` only iterates through the class hierarchy, ignoring any included modules. Would it be possible to support modules here, and if so would there be any objections against adding this?

kernel/common/enumerable.rb
fmfdias: I think the reason at the time was to 'cache' the value and not having to call the [enumerator_size](https://github.com/rubinius/rubinius/blob/master/kernel/common/enumerable.rb#L117-L119) method more times than actually needed.
I didn't measure it, but probably the impact is negligible vs the code readability.
What do you think?

kernel/common/enumerable.rb
fmfdias: The same as above :)

kernel/common/kernel.rb
fmfdias: The to_enum(:loop).size will always return nil unless you give him the size in the block. It's the way [to_enum works](http://ruby-doc.org/core-2.2.0/Object.html#method-i-to_enum). It's up to the developer using the to_enum to provide a size calculation, whenever possible.
You can even do something like this:
```
irb(main):001:0> to_enum(:loop) {"XXL"}.size
=> "XXL"
```
hehehe

fmfdias:
Hi @YorickPeterse,

Sorry about those merges/comments and thanks for pointing the way. I'll do the rebase to clean them up :)

Also thanks for the mirrors insight link. I will read it in more detail and I'll apply them whenever possible.

About the Lazy Enumerators, they are passing the specs. I don't expect the changes made to have any impact on Lazy Enumerators. Like count, the size returns the enumerator size but, unlike it doesn't iterate it to get the value.
When lazy is called the size is passed to the lazy enumerator, but no iteration is done.
E.g:
```
irb(main):001:0> [1, 2].cycle.lazy.collect {|i| p i; i}.size
=> Infinity
```

I'll implement the changes during the following days as time permits.
Thank you!

kernel/common/enumerable.rb
YorickPeterse: Hm, that's a good point actually. I ran a quick benchmark (https://gist.github.com/YorickPeterse/da524a182febcd130436) which outputs the following:

```
Calculating -------------------------------------
          #each_cons    26.586k i/100ms
 #each_cons_no_cache    36.156k i/100ms
-------------------------------------------------
          #each_cons    480.690k (± 6.2%) i/s -      2.393M
 #each_cons_no_cache    448.608k (± 0.6%) i/s -      2.242M

Comparison:
          #each_cons:   480690.2 i/s
 #each_cons_no_cache:   448608.5 i/s - 1.07x slower
```

This shows that it indeed is a little bit faster when using a variable assignment, although not that much. Either way, the rationale is now clear to me. Having said that, I do think that perhaps moving the assignment out of the if statement makes it a bit easier to grok (but this could just be me):

```ruby
enum_size = enumerator_size

if enum_size.nil?

elsif ...

end
```

fmfdias: Cool :) Will make the change!

fmfdias:
Hi @YorickPeterse,

Made a couple of changes based on your review.
I moved the Array private methods to it's mirror class and also created a mirror class for Numeric and Range to do the same with them.
I didn't changed Enumerable#__cycle_size__. This method is being used by Enumerable and by Array because both implement cycle. 
Since this __cycle_size__ method is really a function, where it doesn't need 'self' for anything, one alternative option I thought was to create a Enumerable helper module, (e.g: Rubinius::EnumerableHelper) that could have this and other methods in the same situation. It would be something like this https://gist.github.com/fmfdias/b33e0985373961c32296.

About the rebase... I'm thinking on doing it after we close these loose ends. One question though: Just to confirm, you guys prefer all commits to be squashed into one commit with some context about the 'issue' and the implementation, right?

Thank you!

YorickPeterse:
> About the rebase... I'm thinking on doing it after we close these loose ends. One question though: Just to confirm, you guys prefer all commits to be squashed into one commit with some context about the 'issue' and the implementation, right?

No, it's actually preferred _not_ to squash all of them into a single commit as this leads to massive commits making it harder to cherry-pick things (this was mainly relevant when we were still working on RubySpec). Typically the best approach is to squash commits together only if they are related. For example, you might have one commit that adds something and then another that fixes a typo in said code. Those two commits could be squashed together as there's no real point in keeping them separate.

An example can be found at https://github.com/rubysl/rubysl-socket/pull/9/commits. Initially those commits were 50-something commits which I've merged together as a lot of them were "Added X" commits followed by "Oops, fixed X".

> Since this cycle_size method is really a function, where it doesn't need 'self' for anything, one alternative option I thought was to create a Enumerable helper module, (e.g: Rubinius::EnumerableHelper) that could have this and other methods in the same situation. It would be something like this https://gist.github.com/fmfdias/b33e0985373961c32296.

In Rubinius 3 we'll (most likely) have actual functions for this, solving this problem. For now I think moving it into a helper module is probably best as we don't support `Module` mirrors at the moment.

fmfdias:
Hi @YorickPeterse,

Thanks for the clarifications!
I've implemented the EnumerableHelper module with the cycle_size function to use on both Enumerable and Array also made the rebase.

I ran the specs with ruby 2.2.1 and they also pass, so I think this work can also be reflected on 2.2 branch too.

Ruby 2.2 brought two new Enumerable methods, slice_before and slice_when.
I checked and they both return an enumerator but it's size is nil. There's already a spec for slice_before, that I'll update, but not one for slice_when.
Speaking of slice_when, he's not referenced on #3264. It doesn't exist on the NEWS file referenced in the issus, but it's not the [2.2.0 final release NEWS file](https://github.com/ruby/ruby/blob/v2_2_0/NEWS).

Thank you!

tmornini:
@fmfdias Thanks for all of this work, and @YorickPeterse thanks for coaching it.

Rubinius is so exciting, and it's magical to watch the community working together so effectively.

YorickPeterse:
@fmfdias Specs for slice_after were added in c3fd4a398163d1411da6b1d31bac762a3df1c932, not sure if anybody is currently working on slice_when. I've updated #3264 to also include slice_when. I'll do another run-through of this PR later today/tomorrow, and thanks again for the wonderful work :)

fmfdias:
Hi @tmornini and @YorickPeterse,

Thanks :blush: I'm really glad I could to help! It's been a great experience and I'm enjoying it a lot. :smiley:
Rubinius is indeed a great project and I found it very accessible and easy to participate.
Congrats to you all for that :smiley:!

@YorickPeterse I'll try to do any change request as soon as I can so this one can be released.

Speaking of the #3269, is/was there any similar issue opened for Ruby 2.0/2.1? I'm asking this because while developing this one I found out that on MRI > 2.1, Numeric#step supports keyword arguments. There might be others in similar situation, so I was thinking on checking for this changes that might be missing from Rubinius.
Later I'll open the issue / pull request with specs (and probably an implementation) for the Numeric#spec keyword args.

You guys have a nice weekend!

YorickPeterse:
> Speaking of the #3269, is/was there any similar issue opened for Ruby 2.0/2.1? I'm asking this because while developing this one I found out that on MRI > 2.1, Numeric#step supports keyword arguments. There might be others in similar situation, so I was thinking on checking for this changes that might be missing from Rubinius.

Not sure if I'm following. If you're referring to other changes required for 2.2 compatibility, sure, there's probably stuff not mentioned in the NEWS file and the corresponding issue.

fmfdias:
Sorry... wrong ticket :sweat_smile: The one I wanted to mention was #3264. 

Having a check list is cool to check how much specs are still missing. I was trying to know if there was a similar one for 2.0 and 2.1 compatibility. If they exist, my idea is to compare them to the NEWS file and spot any other missing stuff, like the [Numeric#step keyword arguments](https://github.com/ruby/ruby/blob/v2_1_0/NEWS#L102-L106).

Thanks

YorickPeterse:
As far as I'm aware there's no ticket for what remains to be added for 2.0/2.1.

fmfdias:
Hi @YorickPeterse 

Thanks! I'll try to do a quick check to see if anything more is missing.
In the meantime, I started to add specs and implement the kw args on Numeric#step. I'm using this PR as the base for that work to avoid conflicts.
If I finish it on time and If you agree, I'll merge these specs and implementation to this PR , otherwise I'll create it's own PR.

Thank you!

YorickPeterse:
@fmfdias Doing that in a separate PR is probably best (keeps things clear). However, you can simply branch said PR from this one which should prevent any merge conflicts.

fmfdias:
Hi @YorickPeterse 

Sorry for the delay, but didn't had time to finish the Numeric#step implementation before.
I followed your suggestion and created a branch based on this PR.

Thank you!

YorickPeterse:
Looking into this at the moment, in particular:

* I'm wrapping the Git commits so the lines are at most 80 characters long (http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html is a good read on this). I don't think we mention anything about this anywhere, so don't worry about it.
* It seems some 2.2 specific stuff is in this PR, I'll probably pull that out and merge that into the 2.2 branch instead (if possible). Until we merge the 2.2 branch it's preferred for 2.2 changes to go in a separate PR

Otherwise things are looking fine, I should be done with this in 30 minutes or so :)

fmfdias:
Cool!

Sorry about the commit message. I'll take that into consideration next time I commit to the project :)
About the 2.2 stuff... I tested everything with ruby 2.1.5 and tried only to implement things related to MRI 2.1. Can you point out the stuff you found so I can take a look?

Thank you

YorickPeterse:
The Enum stuff has been rebased into the following commits: https://github.com/rubinius/rubinius/compare/ac5f92af7c86...64d73f9bd5fb

The String#sub changes have been rebased into the 2.2 branch as the following commits: https://github.com/rubinius/rubinius/compare/ccf168676e5b...33dae72abdda

YorickPeterse:
@fmfdias 

> About the 2.2 stuff... I tested everything with ruby 2.1.5 and tried only to implement things related to MRI 2.1. Can you point out the stuff you found so I can take a look?

The commits in https://github.com/rubinius/rubinius/compare/ccf168676e5b...33dae72abdda mentioned things changed in Ruby 2.2. Since we're dropping the use of version guards this means that until we fully support 2.2 changes for said version have to go in a separate branch. Or do the commits mean that this has changed _since_ 2.1 (and not 2.2)?

YorickPeterse:
Ah I see, this changed in 2.1 and not 2.2. In that case I'll merge the commits into the master branch as well.

YorickPeterse:
With this merged/sorted out I'm closing this. Big props for taking care of this! :+1: :beer:

fmfdias:
That behavior comes at least from 2.0.
But 2.1 also introduced some slight behavior changes (Numeric#step TypeError to ArgumentError) :/

fmfdias:
@YorickPeterse 
Thank you! Glad I could help :smile:

PR #3323, opened 2015-02-20
Updating the error message to be equal to the error in ruby variable.c

bjfish:
for compatibility

This causes one error in this build:
https://travis-ci.org/tenderlove/psych/jobs/50729645

Differs from:
https://github.com/ruby/ruby/blob/4194ca02fb2899ce00e6284953d1fd4ee8bdbf38/variable.c

Failing test:
https://github.com/ruby/ruby/blob/eeb05e8c119f8cab6434d90f21551b6bb2954778/test/psych/test_psych.rb

### DIFF
--- a/vm/capi/class.cpp
+++ b/vm/capi/class.cpp
@@ -105,7 +105,7 @@ extern "C" {
 
       if(reason != vFound) {
         free(pathd);
-        rb_raise(rb_eArgError, "undefined class or module %s", path);
+        rb_raise(rb_eArgError, "undefined class/module %s", path);
       }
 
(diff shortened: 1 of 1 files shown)


YorickPeterse:
Is it known why Psych is depending on the exact error message? Generally this isn't part of the compatibility layer, even MRI changes these between minor releases at times.

brixen:
@bjfish please send a PR to that project to fix their dependency on verbatim MRI error messages. There are many reasons not to depend on the exact text of an error message. The proper solution for exception objects is to add accessors that provide the information needed.

Rubinius needs to be free to improve the quality of the error message text. Code should use proper language facilities and not try parsing text.

PR #3343, opened 2015-03-05
Add spec for Enumerable#max optional arg

jsyeo:
I added the spec for new method and cleaned up the specs a bit.


YorickPeterse:
Keep them coming! :+1:

PR #3344, opened 2015-03-05
Fix Mirror.subject= assignments.

thedarkone:
This fixes a common Ruby gotcha.

Besides `Mirror.subject` is not used anywhere, should we remove it?

PS: additionally, would a patch like this get accepted?
```diff
--- a/kernel/bootstrap/mirror.rb
+++ b/kernel/bootstrap/mirror.rb
@@ -1,11 +1,7 @@
 module Rubinius
   class Mirror
-    def self.subject=(klass)
-      @subject = klass
-    end
-
-    def self.subject
-      @subject
+    class << self
+      attr_accessor :subject
     end
 
     def self.reflect(obj)
```

### DIFF
--- a/kernel/bootstrap/array_mirror.rb
+++ b/kernel/bootstrap/array_mirror.rb
@@ -1,7 +1,7 @@
 module Rubinius
   class Mirror
     class Array < Mirror
-      subject = ::Array
+      self.subject = ::Array
 
       def self.reflect(object)
         m = super(object)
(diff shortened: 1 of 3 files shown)


YorickPeterse:
I can't seem to find any references to `subject` in both the `kernel/` and `vm/` directories. My guess is that this is a remnant of an earlier implementation, or it was intended to be used for something which hasn't been implemented (yet). Since I'm not 100% sure I'll let @brixen decide.

YorickPeterse:
Rebased into 798d13ec59c06ac5febbd9b09b2151c23b354961.

PR #3347, opened 2015-03-06
Add specs for Enumerable#slice_when

jsyeo:
:sweat_smile:

### DIFF
--- a/spec/ruby/core/enumerable/slice_when_spec.rb
+++ b/spec/ruby/core/enumerable/slice_when_spec.rb
@@ -1,0 +1,36 @@
+require File.expand_path('../../../spec_helper', __FILE__)
+require File.expand_path('../fixtures/classes', __FILE__)
+
+describe "Enumerable#slice_when" do
+  before :each do
+    ary = [10, 9, 7, 6, 4, 3, 2, 1]
+    @enum = EnumerableSpecs::Numerous.new *ary
+    @result = @enum.slice_when { |i, j| i - 1 != j }
+    @enum_length = ary.length
+  end
+
+  context "when given a block" do
+    it "returns an enumerator" do
+      @result.should be_an_instance_of(enumerator_class)
+    end
+
+    it "splits chunks between adjacent elements i and j where the block returns true" do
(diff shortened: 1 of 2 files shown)


spec/ruby/core/enumerable/slice_when_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  context "when given a block" do
+    it "returns an enumerator" do
+      @result.should be_an_instance_of(enumerator_class)
+    end
+
+    it "splits chunks between adjacent elements i and j where the block returns true" do
+      @result.to_a.should == [[10, 9], [7, 6], [4, 3, 2, 1]]
+    end
+
+    it "calls the block the length of the receiver enumerable minus one times" do
```
YorickPeterse: What is "calls the block the length" supposed to mean?

spec/ruby/core/enumerable/slice_when_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      times_called = 0
+      @enum.slice_when do
+        times_called += 1
+        false
```
YorickPeterse: Any particular reason why this block is returning `false`?

spec/ruby/core/enumerable/slice_when_spec.rb
jsyeo: Should be calls the block for ... times :confounded: . I wasn't particularly awake at the time I wrote it. I have fixed it in the latest commit.

spec/ruby/core/enumerable/slice_when_spec.rb
jsyeo: Alright, I think it's a bad idea to use `false`. I've decided to stick to my `i - 1 != j` example in my latest commit.

YorickPeterse:
Github is now showing 2 commits with the same message, both (as it appears) modifying the same file. Could you rebase these commits into a single one?

jsyeo:
:confused: not sure how that happened. I probably edited the commit message instead of squashing it. Anyway, I've fixed it.

YorickPeterse:
:+1:

PR #3346, opened 2015-03-06
Add specs for Enumerable#max_by optional arg

jsyeo:
:raised_hands:

PR #3348, opened 2015-03-07
Argument error -> ArgumentError

jsyeo:
As per https://github.com/rubinius/rubinius/pull/3346#discussion_r25988583. Just to be consistent.


YorickPeterse:
:+1:

PR #3351, opened 2015-03-08
Implement Enumerable#slice_after

jsyeo:
Now that we have the specs for Enumerable#slice_after, we can implement it!

I hope I didn't confuse anyone with the crazy case when usage. Correct me if I am wrong, but I think it's more readable than a bunch of `if`s and `nil?`s.

### DIFF
--- a/kernel/common/enumerable.rb
+++ b/kernel/common/enumerable.rb
@@ -160,6 +160,31 @@ module Enumerable
     end
   end
 
+  def slice_after(pattern = undefined, &block)
+    pattern_given = !undefined.equal?(pattern)
+
+    raise ArgumentError, "cannot pass both pattern and block" if pattern_given && block_given?
+    raise ArgumentError, "wrong number of arguments (0 for 1)" if !pattern_given && !block_given?
+
+    block = Proc.new { |elem| pattern === elem } if pattern_given
+
+    Enumerator.new do |yielder|
+      accumulator = nil
+      each do |elem|
+        end_chunk = block.yield(elem)
+        accumulator ||= []
+        if end_chunk
+          accumulator << elem
+          yielder.yield accumulator
+          accumulator = nil
+        else
+          accumulator << elem
+        end
+      end
+      yielder.yield accumulator if accumulator
+    end
+  end
+
   def to_a(*arg)
     ary = []
     each(*arg) do


kernel/common/enumerable.rb
```diff
@@ -160,6 +160,33 @@ def slice_before(arg = undefined, &block)
...
+  def slice_after(pattern = undefined, &block)
+    pattern_given = !(undefined.equal? pattern)
+    case [pattern_given, block_given?]
+      when [true, true]
+        raise ArgumentError, "cannot pass both pattern and block"
+      when [false, false]
+        raise ArgumentError, "wrong number of arguments (0 for 1)"
+      when [true, false]
+        block = Proc.new { |elem| pattern === elem }
+    end
```
jemc: Might I suggest the following instead:
```ruby
if pattern_given && block_given?
  raise ArgumentError, "cannot pass both pattern and block"
elsif pattern_given
  block = Proc.new { |elem| pattern === elem }
elsif block_given?
  # use given block
else
  raise ArgumentError, "wrong number of arguments (0 for 1)"
end
```

I think we should avoid allocating a bunch of unnecessary Arrays if we can (it seems like it would be a performance hit), and I think my example is still just as readable, if not a little more so.  The example I gave could be collapsed a bit more into nested `if`s, but I care less about that difference.

jsyeo: Alright. How about this:
```
    raise ArgumentError, "cannot pass both pattern and block" if pattern_given && block_given?
    raise ArgumentError, "wrong number of arguments (0 for 1)" if !pattern_given && !block_given?
    block = Proc.new { |elem| pattern === elem } if pattern_given
    # rest of code
```

kernel/common/enumerable.rb
```diff
@@ -160,6 +160,28 @@ def slice_before(arg = undefined, &block)
     end
   end
 
+  def slice_after(pattern = undefined, &block)
+    pattern_given = !(undefined.equal? pattern)
```
YorickPeterse: `!undefined.equal?(pattern)` is more natural.

kernel/common/enumerable.rb
```diff
@@ -160,6 +160,28 @@ def slice_before(arg = undefined, &block)
...
+    raise ArgumentError, "cannot pass both pattern and block" if pattern_given && block_given?
+    raise ArgumentError, "wrong number of arguments (0 for 1)" if !pattern_given && !block_given?
+    block = Proc.new { |elem| pattern === elem } if pattern_given
+    Enumerator.new do |yielder|
```
YorickPeterse: Some whitespace between these lines would help make it a bit more readable.

kernel/common/enumerable.rb
```diff
@@ -160,6 +160,29 @@ def slice_before(arg = undefined, &block)
     end
   end
 
+  def slice_after(pattern = undefined, &block)
+    pattern_given = !undefined.equal?(pattern)
+    raise ArgumentError, "cannot pass both pattern and block" if pattern_given && block_given?
+    raise ArgumentError, "wrong number of arguments (0 for 1)" if !pattern_given && !block_given?
+    block = Proc.new { |elem| pattern === elem } if pattern_given
```
YorickPeterse: Probably should've been more clear with this (my bad), I meant whitespace between every line, not just the one preceding `Enumerator.new do ...`

jsyeo: Ahh... okay got it. :+1:

jsyeo:
@YorickPeterse I've added the newlines. :smile:

PR #3352, opened 2015-03-10
Implement Binding#local_variables and Binding#receiver

ruipserra:
Hi everyone, this PR implements Binding#local_variables and Binding#receiver.

### DIFF
--- a/kernel/common/binding.rb
+++ b/kernel/common/binding.rb
@@ -54,4 +54,12 @@ class Binding
 
     Kernel.eval(expr, self, filename, lineno)
   end
+
+  def local_variables
+    eval('local_variables')
+  end
+
+  def receiver
+    @self
+  end
 end
--- a/spec/tags/ruby/core/binding/local_variables_tags.txt
+++ b/spec/tags/ruby/core/binding/local_variables_tags.txt
@@ -1,5 +1,0 @@
-fails:Binding#local_variables returns an Array
-fails:Binding#local_variables includes local variables in the current scope
-fails:Binding#local_variables includes local variables defined after calling binding.local_variables
-fails:Binding#local_variables includes local variables of inherited scopes and eval'ed context
-fails:Binding#local_variables includes shadowed local variables only once
--- a/spec/tags/ruby/core/binding/receiver_tags.txt
+++ b/spec/tags/ruby/core/binding/receiver_tags.txt
@@ -1,1 +1,0 @@
-fails:Binding#receiver returns the object to which binding is bound


kernel/common/binding.rb
```diff
@@ -54,4 +54,12 @@ def eval(expr, filename=nil, lineno=nil)
 
     Kernel.eval(expr, self, filename, lineno)
   end
+
+  def local_variables
+    eval('local_variables')
```
YorickPeterse: Euh, we shouldn't need (nor want) the use of `eval` here. If I'm not mistaken the code currently used for `Kernel.local_variables` should be usable in this context, either in the form of a copy-paste (meh) or by using a helper method (one that takes a `Rubinius::VariableScope`).

kernel/common/binding.rb
```diff
@@ -54,4 +54,12 @@ def eval(expr, filename=nil, lineno=nil)
...
+  end
+
+  def receiver
+    @self
```
YorickPeterse: Where does `@self` originate from? Is this not supposed to just return `self` (minus the `@`)?

YorickPeterse:
@jemc I'm reverting this, see my notes.

kernel/common/binding.rb
jemc: It should indeed be `@self` here - `self` would return the `Binding` instance itself.

YorickPeterse: Ah, I now see this is set by `Rubinius::VariableScope#self=`, which wasn't entirely obvious to find.

jemc: To make it more obvious for others in the future, it seems like the `@self` ivar could be renamed to `@receiver`, so that `Binding#receiver` would just be an `attr_reader`.

ruipserra: @jemc I agree. The ```attr_accessor :self``` could be replaced with regular
```def self=(recv) ...``` and ```def self ...``` methods refering to ```@receiver```, which would make the purpose of the ivar clearer, without breaking client code depending on ```self```.

File: kernel/common/binding.rb (hunk 1 above)
Comment (ruipserra) [inline kernel/common/binding.rb]: @YorickPeterse Indeed, the use of eval is a bit gross. The reason behind it was only to avoid duplicating ```Kernel.local_variables```. Would pushing this logic into ```Rubinius::VariableScope``` be an acceptable option?

Comment (YorickPeterse) [inline kernel/common/binding.rb]: Yes, from the top of my head I think this can be reduced down to the following:

```
def local_variables
  scope = Rubinius::VariableScope.of_sender
  
  scope.local_variables
end
```

This can then be used in both `Kernel#local_variables` and `Binding#local_variables`.

jemc: `Binding` already has its `VariableScope` - it's in the `@variables` ivar, from `Binding.setup`.

ruipserra:
Continued at #3354

PR #3354, opened 2015-03-11
Implement Binding#local_variables and Binding#receiver

ruipserra:
This is a follow-up of #3352.

In summary:
* ```Binding#receiver```: I took @jemc's advice and renamed the ```@self``` ivar to ```@receiver```. To avoid breaking other code, ```self```is kept as a virtual attribute.
* ```Binding#local_variables```: Created the ```local_variables```method in ```Rubinius::VariableScope```, which is used both in ```Kernel.local_variables``` and ```Binding#local_variables```.

Comments and feedback welcome. Thanks for your time :smile:

### DIFF
--- a/kernel/common/binding.rb
+++ b/kernel/common/binding.rb
@@ -3,8 +3,16 @@ class Binding
   attr_accessor :compiled_code
   attr_accessor :constant_scope
   attr_accessor :proc_environment
-  attr_accessor :self
   attr_accessor :location
+  attr_reader   :receiver
+
+  def self=(obj)
+    @receiver = obj
+  end
+
+  def self
+    @receiver
+  end
 
   def from_proc?
     @proc_environment
@@ -53,5 +61,9 @@ class Binding
     lineno ||= filename ? 1 : line_number
 
     Kernel.eval(expr, self, filename, lineno)
+  end
+
+  def local_variables
+    variables.local_variables
   end
 end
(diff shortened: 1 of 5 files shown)


kernel/common/binding.rb
```diff
@@ -3,8 +3,16 @@ class Binding
...
   attr_accessor :location
+  attr_reader   :receiver
+
+  def self=(obj)
+    @receiver = obj
+  end
+
+  def self
+    @receiver
+  end
```
jemc: Do we even need to keep `#self` and `#self=` around?

If it's not part of a Ruby-language-compatibility contract, and only Rubinius internal code uses these, can we just make them `attr_accessor :receiver` instead?

YorickPeterse:
This has been merged into 59d6f28ce534cc4d75736dbdac70b83d458f476c (had to do a rebase first to keep the logs clean, so GH didn't notice).

jemc:
I guess I'll see about the `self` cleanup myself then and do it in another commit.

kernel/common/binding.rb
YorickPeterse: @jemc Wouldn't that make it set `@self` again instead?

ruipserra:
Thanks! @jemc You're right that ```#self``` and ```#self=``` aren't really needed, but I didn't want to start ripping things apart without first checking with you guys.

jemc:
@ruipserra 
No worries, thanks for working on this!

jemc:
https://github.com/rubinius/rubinius/commit/ccf168676e5bf1d4e8dba19cf20abdaa42170d28

PR #3355, opened 2015-03-11
Allow shared linking to LLVM 3.5

heftig:
Run check_llvm_flags before checking for shared libs and use the
calculated flags there.

### DIFF
--- a/configure
+++ b/configure
@@ -563,15 +563,6 @@ class Configure
       return false
     end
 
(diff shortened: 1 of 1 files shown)


YorickPeterse:
Rebased into 4fd2b3dc1395e838d5e77de4b551063d5849deaa, closing manually since GH is too dumb to detect this. Thanks!

PR #3356, opened 2015-03-12
Add numeric step kw args

fmfdias:
Hi guys,

This PR adds support for keyword arguments on Numeric#step method.
Since this method spec and implementation were changed during the implementation of #3320, these changes were made over that to avoid conflicts. Because of this, this PR should only be merged after #3320.

Thank you!


YorickPeterse:
With #3320 being merged, could you rebase according to the master branch? That should get rid of the enum commits from this PR.

fmfdias:
Rebase done!

YorickPeterse:
Looking at this, is this based on recent changes in MRI? If so, is there any link to the commit(s)/NEWS file detailing these changes?

spec/ruby/core/numeric/shared/step.rb
```diff
@@ -0,0 +1,409 @@
+require File.expand_path('../../../../spec_helper', __FILE__)
+require File.expand_path('../../fixtures/classes', __FILE__)
+
+describe :numeric_step, :shared => true do
+
+  def get_args(*args)
```
YorickPeterse: I'm not sure if this was already present before (bit hard to grok based on the diff), but if not then this should be moved into a fixture file of sorts. Doing so keeps the spec files clean.

fmfdias:
This is based on MRI 2.1. 
Here's the NEWS lines https://github.com/ruby/ruby/blob/v2_1_0/NEWS#L102-L106

spec/ruby/core/numeric/shared/step.rb
fmfdias: The get_args method is new and it's specific to step method specs with kw arguments. It allows to run the tests with different argument passing strategies.
My idea would be to move it into a into a specific helper file. What do you think?

fmfdias:
Hi @YorickPeterse,

Moved the get_args method to a helper file, but I'm still not sure if it's the best place for it. 
Should it go to the numeric helper on mspec?

Thanks

YorickPeterse:
@fmfdias We might not need it at all to begin with, but I'll have to dig through the specs myself to see if that is the case. I'll try to take a look at it this weekend.

fmfdias:
Ok :) Let me know of any changes needed!

kernel/common/numeric.rb
```diff
@@ -31,23 +31,27 @@ def <=>(other)
...
+  def step(limit=nil, step=nil, to: nil, by: nil)
     unless block_given?
-      return to_enum(:step, limit, step) do
-        Rubinius::Mirror::Numeric.reflect(self).step_size(limit, step)
+      return to_enum(:step, limit, step, to: to, by: by) do
+        Rubinius::Mirror::Numeric.reflect(self).step_size(limit, step, to, by)
       end
     end
 
-    raise ArgumentError, "step cannot be 0" if step == 0
```
YorickPeterse: This error is still raised in case of the following:

    10.step(100, 0).to_a => ArgumentError: step can't be 0

kernel/common/numeric.rb
```diff
@@ -31,23 +31,27 @@ def <=>(other)
...
+    values = m.step_fetch_args(limit, step, to, by)
     value = values[0]
     limit = values[1]
     step = values[2]
     asc = values[3]
     is_float = values[4]
 
+    if(step == 0)
+      while true
+        yield self
```
YorickPeterse: How is this supposed to ever return from the method (other than the supplied block calling `break`)?

kernel/common/numeric.rb
```diff
@@ -31,23 +31,27 @@ def <=>(other)
...
     m = Rubinius::Mirror::Numeric.reflect(self)
-    values = m.step_fetch_args(limit, step)
+    values = m.step_fetch_args(limit, step, to, by)
     value = values[0]
     limit = values[1]
     step = values[2]
     asc = values[3]
     is_float = values[4]
 
+    if(step == 0)
```
YorickPeterse: Use `if step == 0`, the parenthesis are not required here.

spec/ruby/core/numeric/helpers/step.rb
```diff
@@ -0,0 +1,13 @@
+class Object
+  def get_args(args_type, *args)
```
YorickPeterse: Thinking of it, the specs should probably be refactored to remove the need for this helper method. Right now it's hard to see what exactly the specs are doing as they rely on this rather odd helper method.

spec/ruby/core/numeric/shared/step.rb
```diff
@@ -0,0 +1,397 @@
...
+
+  it "returns an Enumerator when step is 0" do
+    1.send(@method, *get_args(@args_type, 2, 0)).should be_an_instance_of(enumerator_class)
+  end
+
+  it "returns an Enumerator when not passed a block and self > stop" do
+    1.send(@method, *get_args(@args_type, 0, 2)).should be_an_instance_of(enumerator_class)
+  end
+
+  it "returns an Enumerator when not passed a block and self < stop" do
```
YorickPeterse: This and the above spec should probably be wrapped in another `describe` like so as this makes them a bit easier to grok (at least in my opinion):

```ruby
describe 'with a block' do
  it 'returns an Enumerator when step is 0' do

  end

  ...
end
```

spec/ruby/core/numeric/shared/step.rb
```diff
@@ -0,0 +1,397 @@
...
+    1.send(@method, *get_args(@args_type, 2, 3)).should be_an_instance_of(enumerator_class)
+  end
+
+  it "returns an Enumerator that uses the given step" do
+    0.send(@method, *get_args(@args_type, 5, 2)).to_a.should == [0, 2, 4]
+  end
+
+  describe "with [stop, step]" do
```
YorickPeterse: Use regular English here, so `with stop and step` or something along those lines.

spec/ruby/core/numeric/shared/step.rb
```diff
@@ -0,0 +1,397 @@
...
+    end
+
+    it "increments self using #+ until self > stop when step > 0" do
+      @step.should_receive(:>).with(0).and_return(true)
+      @obj.should_receive(:>).with(@stop).and_return(false, false, false, true)
+      @obj.should_receive(:+).with(@step).and_return(@obj, @obj, @obj)
+
+      @obj.send(@method, *get_args(@args_type, @stop, @step), &@prc)
+
+      ScratchPad.recorded.should == [@obj, @obj, @obj]
```
YorickPeterse: These tests don't test behaviour, they test implementation details. Specs shouldn't dictate what methods internally are used except for a few rare edge cases where calling a certain method is part of the defined behaviour. In this case it's crazy to test that the objects called `>` and `+`.

spec/ruby/core/numeric/shared/step.rb
```diff
@@ -0,0 +1,397 @@
...
+    end
+  end
+
+  describe "Numeric#step with [stop, step] when self, stop and step are Fixnums" do
+    it "yields only Fixnums" do
+      1.send(@method, *get_args(@args_type, 5, 1)) { |x| x.should be_kind_of(Fixnum) }
+    end
+  end
+
+  describe "Numeric#step with [stop, step] when self and stop are Fixnums but step is a String" do
```
YorickPeterse: This describe block, along with the one above, should be broken up into separate blocks. For example:

```
describe 'Numeric#step' do
  describe 'with stop, step and self as a Fixnum' do
    it 'yields only Fixnums' do
      ...
    end
  end

  describe 'with stop and self as a Fixnum and step as a String' do
    it 'returns an Enumerator if not given a block' do
      ...
    end
  end
end
```

While this introduces some extra indentation levels it cuts down the length of each individual spec/describe description. It also groups together related specs instead of them being toplevel `it` blocks.

spec/ruby/core/numeric/shared/step.rb
```diff
@@ -0,0 +1,397 @@
...
+    @prc = lambda { |x| ScratchPad << x }
+  end
+
+  it "defaults to step = 1" do
+    1.send(@method, *get_args(@args_type, 5), &@prc)
+    ScratchPad.recorded.should == [1, 2, 3, 4, 5]
+  end
+
+  it "returns an Enumerator when step is 0" do
+    1.send(@method, *get_args(@args_type, 2, 0)).should be_an_instance_of(enumerator_class)
```
YorickPeterse: Where does `enumerator_class` come from?

spec/ruby/core/numeric/shared/step.rb
```diff
@@ -0,0 +1,397 @@
...
+      1.send(@method, *get_args(@args_type, 5, "0.1")).should be_an_instance_of(enumerator_class)
+      1.send(@method, *get_args(@args_type, 5, "1/3")).should be_an_instance_of(enumerator_class)
+      1.send(@method, *get_args(@args_type, 5, "foo")).should be_an_instance_of(enumerator_class)
+    end
+
+    it "raises an ArgumentError if given a block" do
+      lambda { 1.send(@method, *get_args(@args_type, 5, "1")) {} }.should raise_error(ArgumentError)
+      lambda { 1.send(@method, *get_args(@args_type, 5, "0.1")) {} }.should raise_error(ArgumentError)
+      lambda { 1.send(@method, *get_args(@args_type, 5, "1/3")) {} }.should raise_error(ArgumentError)
+      lambda { 1.send(@method, *get_args(@args_type, 5, "foo")) {} }.should raise_error(ArgumentError)
```
YorickPeterse: All these expectations should be moved into separate `it` blocks, same goes for the specs below. For example:

```ruby
describe 'with a block' do
  it 'raises ArgumentError with step "0.1"' do
    
  end

  it 'raises ArgumentError with step "1/3"' do
    
  end

  ...
end
```

spec/ruby/core/numeric/shared/step.rb
```diff
@@ -0,0 +1,397 @@
...
+
+    it "does not yield when self is less than stop" do
+      1.send(@method, *get_args(@args_type, 5, -1), &@prc)
+      ScratchPad.recorded.should == []
+    end
+  end
+
+  describe "Numeric#step with [stop, step]" do
+    it "yields only Floats when self is a Float" do
+      1.5.send(@method, *get_args(@args_type, 5, 1)) { |x| x.should be_kind_of(Float) }
```
YorickPeterse: Use `should be_an_instance_of` instead as we want to test if the object is an actual instance.

YorickPeterse:
Regarding some of the comments I made: I know part of these specs were copied over from the existing file, but I think now is a good time to clean them up properly.

kernel/common/numeric.rb
fmfdias: The logic moved to the mirror since to allow it to be shared with the part where the enumerator size is calculated.
You can see it in Rubinius::Mirror::Numeric#step_fetch_args.

kernel/common/numeric.rb
fmfdias: :+1:

kernel/common/numeric.rb
fmfdias: At far as I could see, it's not supposed too. It's how MRI behaves... If step, given by the by: keyword argument, is 0, you get an infinite loop.
From the [Numeric#step documentation](http://ruby-doc.org/core-2.1.5/Numeric.html#method-i-step):
> The loop finishes when the value to be passed to the block is greater than limit (if step is positive) or less than limit (if step is negative), where limit is defaulted to infinity.

You can still do stuff like
```
10.step(to: 15, by: 0).take(5)
=> [10, 10, 10, 10, 10]
```
... although I don't immediately see it's usefulness.

spec/ruby/core/numeric/helpers/step.rb
fmfdias: Besides the step = 0 vs by: 0 difference and the ArgumentError raising when repeating arguments when using positional and keyword arguments, the results should always be the same.

The idea behind this was to avoid duplicating the specs 3 times where the only thing that changes is how you pass the arguments.

I can thing on other options, and I'm open to suggestions, being it a different approach or clarifying this method to avoid confusion, but I think that duplicating the tests is the worst one.

spec/ruby/core/numeric/shared/step.rb
fmfdias: It's being used all over the place :)
I think it comes from [here](https://github.com/rubinius/rubinius/blob/master/mspec/lib/mspec/helpers/enumerator_class.rb).

fmfdias:
Hi @YorickPeterse 

Thanks for the review :)
I replied to some of your comments.

About the spec cleaning, I'll gladly help clearing out this specs.
My main objective was to introduce specs for the new behavior without changing too much of specs that don't need to change to fulfill the purpose. I will continue to proceed this way unless you guys ask for it, like in this case :)

Thank you!
Have a nice Sunday!

fmfdias:
Hi @YorickPeterse,

Started the spec cleanup/reorganization.
* Removed the specs that were testing implementations details
* Reorganized the test into groups with similar properties
* Tried to improve their texts (I think they may still need some improvement)
* Changed from be_kind_of to be_an_instance_of

In terms of the "step as string" argument error specs, I think the idea behind the tests that are being made is to check that even when the strings are numeric representations, the exception is raise. Having this is mind I reorganized them in a slightly different way, but if you think they don't make send, I'll act accordingly :)

While doing this I fixed some specs that could result in false positives (they were using == instead of eql) and fixed one issue that wasn't being caught because of this.

About the ruby_bug guards, should I remove them too?

Don't worry about the commits, I'll rebase this in the end.
Thank you!

YorickPeterse:
> About the ruby_bug guards, should I remove them too?

Yeah, they can be removed. I'll take a closer look this weekend.

fmfdias:
Hi @YorickPeterse 

Cool :)
Removed the ruby_bug guards.

Thank you!

YorickPeterse:
Shit, I completely forgot to take a look at this since. I've noted it down for this weekend instead.

fmfdias:
No problem, hehehe :) I also didn't had much time to help these past few weeks.

YorickPeterse:
Having taken another look, I still stand by having to somehow refactor (out) `Object#get_args`. I'll play around with the changes during the coming week or so to see if I can come up with a nice way to solve this problem.

fmfdias:
Hi @YorickPeterse,

Sorry about the late reply. Unfortunately I'm not having too much time lately. Today I had a little, so I tried to come back to this and try a different approach to the last standing issue :)

Instead of using a helper method as a way help share specs between different argument passing types, the shared spec now requires a variable with a Proc/lambda to transform the arguments to the desired argument passing type. 
When using the shared spec, a lambda that does the transformation is defined for each of the argument type that we want to test.
Most important, to help understand what's being done I added some comments near the lambdas definitions and a comment on the shared spec definition.

While doing this, I also synced master changes into the branch.

Thanks!

YorickPeterse:
I'll take a look at this somewhere this week. 

One thing I wonder: do we have to duplicate tests for both positional and keyword argument forms? Can we perhaps not write the tests for positional arguments and then a few simple ones to check if keyword arguments are also used? I'm not sure if this would be more rigid, but it would probably simplify the sharing of specs a bit.

fmfdias:
Hi @YorickPeterse,

Thanks!
I think I see your point... Since the behavior of the method is the same* no whatever the argument style is, we may skip repeating it for different form types.
The idea behind this is that they are different arguments that have the same meaning, much like having an alias method. In those cases we also test that the alias method behaves like the original one.
To complicate things a bit, we can mix positional and keyword arguments style and this is also being tested.
An example of why this might come in handy: I ran the test against jruby 9000 pre2 and the test enters in a infinite loop in the shared section.

* When using keyword arguments the method accept 0 as step, while the positional argument style doesn't

Thank you!

brixen:
@fmfdias sorry it took so long to merge this. Thanks for the help!

Issue #3360, opened 2015-03-17
Compiler error for splat operator followed by more arguments in case-when


maltoe:
Hi,

not sure if this qualifies as a bug, but at least this code used to work in MRI. Using rubinius 2.5.2.

```ruby
a = [1, 2]
case 3
when 3, *a
  puts 'works'
when *a, 3
  puts 'compiler error, but works in MRI'
end
```

System information:

```sh
$ ruby -v
rubinius 2.5.2 (2.1.0 7a5b05b1 2015-01-30 3.0 JI) [x86_64-linux-gnu]
$ rbenv -v
rbenv 0.4.0-98-g13a474c
$ uname -a
Linux xyz 3.2.0-4-amd64 #1 SMP Debian 3.2.65-1+deb7u1 x86_64 GNU/Linux
```

[Gist with stacktrace](https://gist.github.com/maltoe/b43657ffb88f5f178340)

Otherwise, thanks for this great product!

jemc:
@maltoe 
Thanks for reporting this!  I think I have a fix that works, but I'm not that familiar with this behavior of `case`/`when` and I want to make sure it's done right.  Would you mind writing a few specs in `spec/ruby/language/case_spec.rb` to demonstrate a normal case and some edge cases for the behavior (which you probably understand better than I do at the moment).

maltoe:
In fact, I too have no real clue what is a "normal" case here, I just stumbled across this weird behaviour. Digging into it, it only got more interesting. See the PR, where only the latter two test cases fail. It seems like somehow the variable expansion is messed up when it is followed by additional values (yet for two values it works with a single splat...).

PR #3362, opened 2015-03-17
Add specs for #3360 (case-when with splatted arrays from variables).

maltoe:
See issue #3360.

### DIFF
--- a/spec/ruby/language/case_spec.rb
+++ b/spec/ruby/language/case_spec.rb
@@ -143,6 +143,47 @@ describe "The 'case'-construct" do
     end.should == "foo"
   end
 
+  it "takes an expanded array before additional listed values" do
+    case 'f'
+      when *['a', 'b', 'c', 'd'], 'f'
+        "foo"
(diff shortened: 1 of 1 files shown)


jemc:
Ah, so FYI, normally I would ask that you run `mspec tag fails spec/ruby/language/case_spec.rb` to have `mspec` run through the new tests and tag them as failing so that the CI would still pass.  However, in this case, because it's a compiler error it's impossible to even do that much.  Also, the fact that the bug is in the Rubinius code tools gems makes this a little more difficult to deal with as well.

So here's what I'll do - I'll pull your tests down locally to work on the fix on my dev machine.  I'll push a fix to the `rubinius-ast` gem, then get @brixen to do a release of the gem.  Then we can bump the version of `rubinius-ast` that Rubinius depends on and the issue will be fixed on master, and I can merge these tests in this PR because they'll pass.

jemc:
@maltoe 
Thanks for reporting and for writing some specs!

PR #3364, opened 2015-03-19
Add 2.2 language specs

ruipserra:
This PR adds specs for the the following language changes in ruby 2.2 ([see NEWS file](http://github.com/ruby/ruby/blob/v2_2_0/NEWS)):

* nil/true/false objects are frozen. [Feature #8923]
* Hash literal: Symbol key followed by a colon can be quoted. [Feature #4276]
* fixed a very longstanding bug that an optional argument was not
  accessible in its default value expression.  [Bug #9593]

I'll go over some points about the commits:

* I wasn't sure where to put the nil/true/false specs. I considered creating `nil_spec.rb`, `true_spec.rb` and `false_spec.rb` in `spec/ruby/language` with a single example for `frozen?`, but for the sake of consistency with the existing specs I ended up putting them in `kernel/frozen_spec.rb`. Was this the best choice?
* The expectation for the new quoted symbol syntax uses an  `eval` because otherwise it would break the whole file (the parser would throw a syntax error and no examples would be run).

Thank you for your time.

### DIFF
--- a/kernel/common/binding.rb
+++ b/kernel/common/binding.rb
@@ -54,12 +54,4 @@ class Binding
 
     Kernel.eval(expr, self, filename, lineno)
   end
-
-  def local_variables
-    eval('local_variables')
-  end
-
-  def receiver
-    @self
-  end
 end
--- a/kernel/common/binding.rb
+++ b/kernel/common/binding.rb
@@ -54,4 +54,12 @@ class Binding
 
     Kernel.eval(expr, self, filename, lineno)
   end
+
+  def local_variables
+    eval('local_variables')
+  end
+
+  def receiver
+    @self
+  end
 end
--- a/kernel/common/binding.rb
+++ b/kernel/common/binding.rb
@@ -3,8 +3,16 @@ class Binding
   attr_accessor :compiled_code
   attr_accessor :constant_scope
   attr_accessor :proc_environment
-  attr_accessor :self
   attr_accessor :location
+  attr_reader   :receiver
+
+  def self=(obj)
+    @receiver = obj
+  end
+
+  def self
+    @receiver
+  end
 
   def from_proc?
     @proc_environment
@@ -53,5 +61,9 @@ class Binding
     lineno ||= filename ? 1 : line_number
 
     Kernel.eval(expr, self, filename, lineno)
+  end
+
+  def local_variables
+    variables.local_variables
   end
 end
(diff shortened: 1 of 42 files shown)


YorickPeterse:
> [...] consistency with the existing specs I ended up putting them in kernel/frozen_spec.rb. Was this the best choice?

Looks fine to me.

> The expectation for the new quoted symbol syntax uses an eval because otherwise it would break the whole file (the parser would throw a syntax error and no examples would be run).

We'll probably need to re-generate the parser first, then we should be able to remove the use of `eval`.

ruipserra:
Hey @YorickPeterse, thanks for the feedback. I'll probably write a few more language specs in the next couple of days. I'll update the PR when I do.

ruipserra:
Hi everyone,

* Added block and lambda literal specs for MRI issue [#9593](https://bugs.ruby-lang.org/issues/9593) (optional argument accesible in its default value expression)
* Updated hash_spec according to MRI issue [#10315](https://bugs.ruby-lang.org/issues/10315) ([ChangeLog](https://github.com/ruby/ruby/blob/v2_2_0/ChangeLog#L3153..L3163))
* Removed assertion checking that U+180E matches `/[[:blank:]]/`, as it's no longer considered a whitespace character

At this point I still see 3 failed specs with MRI 2.2.1 in `spec/ruby/language`, but I suspect one or two of those to be MRI bugs. I opened this [issue](https://bugs.ruby-lang.org/issues/10994), and I'm still waiting for feedback.

YorickPeterse:
@ruipserra Any feedback on the Redmine ticket you submitted? Sometimes it can be a bit hard to get a hold of Matz & co, in that case replying to the ticket (and specifically mentioning somebody like Koichi or Nobu) sometimes can do the trick.

Unless this particular Redmine ticket is blocking I'll take a look at this PR this weekend.

tak1n:
Fail, I didn't see this PR until now.
Many of these commits are already on the 2.2 branch.
How should we proceed?

YorickPeterse:
I think this needs a rebase as there are a bunch of commits that aren't yet on our 2.2 branch.

brixen:
The easiest way to do this is probably cherry-pick the commits from here to the 2.2 branch so we can close this and avoid the confusion. If @ruipserra, @YorickPeterse, or @tak1n can't do this, I'll do it in a day or two.

YorickPeterse:
Commits rebased into the 2.2 branch, thanks!

tak1n:
Awesome thanks :)

PR #3372, opened 2015-04-07
Add local var methods to Binding

mlarraz:
These are defined in MRI as of 2.1.

I basically just copied these from the [RubyDocs](http://ruby-doc.org/core-2.1.0/Binding.html#method-i-local_variable_defined-3F).

My implementation of `Binding#local_variable_set` assumes obj responds to #to_s, which may cause issues.

### DIFF
--- a/kernel/common/binding.rb
+++ b/kernel/common/binding.rb
@@ -56,14 +56,14 @@ class Binding
   end
 
   def local_variable_defined?(symbol)
-    eval("defined?(#{symbol}) == 'local-variable'")
+    variables.local_defined?(symbol)
   end
 
   def local_variable_get(symbol)
-    eval("#{symbol}")
+    variables.get_eval_local(symbol)
   end
 
   def local_variable_set(symbol, obj)
-    eval("#{symbol} = #{obj}")
+    variables.set_eval_local(symbol, obj)
   end
 end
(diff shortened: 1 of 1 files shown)


YorickPeterse:
No, we're not going to rely on `eval` for this. Instead we should use `Rubinius::VariableScope` to get/set any local variables in a scope. See https://github.com/rubinius/rubinius/blob/7362475af1247f8f3622851e5d4fd486ff5826a9/kernel/common/eval.rb#L109-L112 as an example.

Also, if we have any tagged specs for this they'll need to be untagged, otherwise new specs should be added.

mlarraz:
So my only question about using `Rubinius::VariableScope` is that the getter/setter methods only seem to consider dynamic locals. Is there an API to get/set locals from `Rubinius::CompiledCode`?

YorickPeterse:
`VariableScope` puts eval'd variables under `VariableScope#dynamic_locals`. Looking at things we probably need to expose some extra bits to this class. There's `VariableScope#set_local` but this requires a slot number (= basically the position in the local variables list) instead of just a variable name.

brixen:
Are there specs for this?

YorickPeterse:
Don't think so, at least I can't find any.

YorickPeterse:
Fixed in 40f04c11393f6314189168bf46eb02893f055bde, thanks for initially looking into this.

PR #3379, opened 2015-04-26
Clean bundler environment before running Homebrew during configuration

d4rky-pl:
Bundler leaks RUBYOPT='-rbundler/setup' to Homebrew which may cause it to crash when system Ruby doesn't have bundler installed. This causes RVM to crash during `rvm install rbx`, see rvm/rvm#3386

### DIFF
--- a/configure
+++ b/configure
@@ -576,7 +576,7 @@ class Configure
         if macports?
           config = macports_llvm_config
         else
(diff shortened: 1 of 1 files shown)


d4rky-pl:
This patch should've nothing to do with why Travis CI build fail. Any way to retry the build without additional commits?

PR #3402, opened 2015-05-16
Minor documentation fixes

benlovell:
I was scanning the `CONTRIBUTING` guide and noticed a couple minor issues.
:heart_eyes:


jemc:
I don't know if I can speak for everyone else, but I personally don't like editing markdown that is 'hard-wrapped'.  Most text editors will soft-wrap it for you, and it saves you the trouble of having to add and remove newlines every time you change a word or sentence.

benlovell:
@jemc fair point, I guess it's a matter of opinion.

YorickPeterse:
@jemc soft wrapping is a tool for the devil, I've yet to find a single editor that can do it well. Also not everybody views Markdown on either Github or in an editor capable of soft wrapping lines. For example, I regularly read text files using `less` which is notoriously bad at this. As such I'd actually prefer hard wrapped lines.

PR #3405, opened 2015-05-20
Fixed FreeBSD 10.x build issue, caused by syscall SYS_gettid which is…

dmilith:
… Linux specific.


brixen:
@dmilith thanks for the working on this patch. I extracted the platform-specific stuff behind a function interface in https://github.com/rubinius/rubinius/commit/05613d25608e10b021469faf6a639ed3d42fe85a.

PR #3410, opened 2015-05-25
Fix a __GNUC__ check to work with GCC 5

heftig:


heftig:
Also, I'm not sure SPECIALIZATION_STORAGE can't still be "inline".

PR #3414, opened 2015-05-28
Implemented File birthtime

tak1n:
Opening this pr because I'm not sure if I did it the "right" way.

The configure script checks if on your system in `stat.h` the function `birthtime()` is defined.
If it is a `HAVE_ST_BIRTHTIME` macro is set which can be used in c++ to check whether to return `st_.st_birthtime` as Time Object or return a `Primitives::failure()`.

In the ruby Stat#birthtime method a `NotImplementedError` is raised if the primitive fails.
Also updated the specs for the specific platforms, the problem here I'm not sure which platform supports `birthtime()` and which don't.

Belongs to #3264

### DIFF
--- a/configure
+++ b/configure
@@ -1231,7 +1231,7 @@ int main() { return tgetnum(""); }
     end
 
     if has_function("birthtime", ["sys/stat.h"])
-       @defines << "HAVE_ST_BIRTHTIME"
+      @defines << "HAVE_ST_BIRTHTIME"
     end
 
     # glibc has useless lchmod() so we don't try to use lchmod() on linux
--- a/configure
+++ b/configure
@@ -1230,7 +1230,7 @@ int main() { return tgetnum(""); }
       @defines << "HAVE_GETTID"
     end
 
-    if has_function("birthtime", ["sys/stat.h"])
+    if has_struct_member("stat", "st_birthtimespec", ["sys/stat.h"])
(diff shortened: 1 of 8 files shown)


chuckremes:
I think this is probably okay to merge. I don't know when I'll finish the ffi-io branch and merge it to master (assuming it's acceptable). In the meantime, this solves this particular gap.

configure
```diff
@@ -1230,6 +1230,10 @@ int main() { return tgetnum(""); }
       @defines << "HAVE_GETTID"
     end
 
+    if has_function("birthtime", ["sys/stat.h"])
+       @defines << "HAVE_ST_BIRTHTIME"
+    end
```
jemc: Sorry to be picky here, but I line 1234 has an extra leading whitespace.

kernel/common/file.rb
```diff
@@ -104,6 +104,28 @@ def self.atime(path)
...
+  # file was changed, not the file itself).
+  #
+  #  File.ctime("testfile")   #=> Wed Apr 09 08:53:13 CDT 2003
+  def self.ctime(path)
+    Stat.new(path).ctime
+  end
+
+  def self.birthtime(path)
+    Stat.new(path).birthtime
+  end
```
jemc: I think this method should have a documentation comment above it as well to be consistent with the other two.

PR #3417, opened 2015-05-30
Implement Float#next_float Float#prev_float

jsyeo:
I followed MRI's C implementation for this. :sweat_smile: Let me know if it's alright.

### DIFF
--- a/kernel/common/float.rb
+++ b/kernel/common/float.rb
@@ -285,4 +285,32 @@ class Float < Numeric
     return int if self > 0 or self == int
     return int - 1
   end
+
+  def next_float
+    return NAN if self.nan?
+    return -MAX if self == -INFINITY
+    return INFINITY if self == MAX
+    return Math.ldexp(0.5, MIN_EXP - MANT_DIG + 1) if self.zero?
+
+    frac, exp = Math.frexp self
+
+    if frac == -0.5
+      frac *= 2
+      exp -= 1
+    end
+
+    smallest_frac = EPSILON / 2
+    smallest_frac = Math.ldexp(smallest_frac, MIN_EXP - exp) if exp < MIN_EXP
+
+    result_frac = frac + smallest_frac
+
+    return -0.0 if result_frac.zero? && frac < 0
+    return 0.0 if result_frac.zero? && frac > 0
+
+    return Math.ldexp result_frac, exp
+  end
+
+  def prev_float
+    return -(-self).next_float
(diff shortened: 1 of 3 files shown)


kernel/common/float.rb
```diff
@@ -285,4 +285,32 @@ def floor
...
+
+  def next_float
+    return NAN if self.nan?
+    return -MAX if self == -INFINITY
+    return INFINITY if self == MAX
+    return Math.ldexp 0.5, MIN_EXP - MANT_DIG + 1 if self.zero?
+
+    frac, exp = Math.frexp self
+
+    if frac == -0.5 then
```
YorickPeterse: No need for `then` here.

kernel/common/float.rb
```diff
@@ -285,4 +285,32 @@ def floor
     return int if self > 0 or self == int
     return int - 1
   end
+
+  def next_float
+    return NAN if self.nan?
+    return -MAX if self == -INFINITY
+    return INFINITY if self == MAX
+    return Math.ldexp 0.5, MIN_EXP - MANT_DIG + 1 if self.zero?
```
YorickPeterse: Explicit parenthesis are preferred here as it would help readability:

    return Math.ldexp(0.5, MIN_EXP - MANT_DIG + 1) if self.zero?

kernel/common/float.rb
```diff
@@ -285,4 +285,32 @@ def floor
...
+
+    frac, exp = Math.frexp self
+
+    if frac == -0.5 then
+      frac *= 2
+      exp -= 1
+    end
+
+    d = EPSILON / 2
+    d = Math.ldexp(d, Float::MIN_EXP - exp) if exp < Float::MIN_EXP
```
YorickPeterse: This can just use `MIN_EXP` since it's in the same namespace.

jsyeo:
@YorickPeterse fixed!

kernel/common/float.rb
```diff
@@ -285,4 +285,32 @@ def floor
...
+
+    if frac == -0.5
+      frac *= 2
+      exp -= 1
+    end
+
+    d = EPSILON / 2
+    d = Math.ldexp(d, MIN_EXP - exp) if exp < MIN_EXP
+
+    result_frac = frac + d
```
YorickPeterse: What does `d` stand for in this case? This is rather picky but I personally despise single letter variables.

PR #3420, opened 2015-06-02
Implement Enumerable#{min,max,min_by,max_by} with optional arguments for Ruby 2.2

kachick:


kernel/common/enumerable.rb
```diff
@@ -567,86 +567,110 @@ def first(n=undefined)
...
-        min = o
-      else
-        comp = block_given? ? yield(o, min) : o <=> min
-        unless comp
-          raise ArgumentError, "comparison of #{o.class} with #{min} failed"
-        end
-
-        if Comparable.compare_int(comp) < 0
+  def min(n=nil)
+    if n.nil?
```
YorickPeterse: I can't really read it from the diff, but is this method supposed to return `nil` if `n` is also `nil`? In that case this is much better:

    return nil if n.nil?

kernel/common/enumerable.rb
```diff
@@ -567,86 +567,110 @@ def first(n=undefined)
     nil
   end
 
-  def min
-    min = undefined
-    each do
-      o = Rubinius.single_block_arg
```
YorickPeterse: What is `o` supposed to mean in this case? Can do get rid of the single letter variables?

kachick: I didn't touch the code, but the variable names have been rewritten :)

YorickPeterse: Oh I wasn't condemning you for it, but when changing this anyway it would be nice to take care of cases like this :)

kachick:
Thanks!

PR #3425, opened 2015-06-05
Enumerator::Lazy#take should switch size with Infinity

kachick:


kernel/common/enumerator.rb
```diff
@@ -258,7 +258,7 @@ def take(n)
         raise ArgumentError, "attempt to take negative size" if n < 0
 
         current_size = enumerator_size
-        set_size = if current_size.kind_of?(Integer)
+        set_size = if current_size.kind_of?(Numeric)
```
YorickPeterse: Not sure if I understand this change, is `current_size` a Float in this case?

kachick: Yes, Infinity(Float::Infinity) is a Float.

```ruby
loop.size.class #=> Float
```

YorickPeterse: Ah yeah, now it makes sense.

YorickPeterse:
Thanks!

kachick:
Thanks to merge!

PR #3432, opened 2015-06-10
Some CAPI fixes

tamird:


jemc:
I would probably merge this for you now if it weren't for the changes to `configure` - I don't know much about OSX or brew so I can't confirm those changes.

tamird:
I promise that `configure` didn't work before the change, and does after the change :)

tamird:
I've updated this to run CI on OSX

.travis.yml
```diff
@@ -7,16 +7,16 @@ compiler:
 before_install:
   - echo $LANG
   - echo $LC_ALL
-  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get update && sudo apt-get install -y llvm-3.4 llvm-3.4-dev; fi
-  - if [ $TRAVIS_OS_NAME == osx ]; then brew update && brew install llvm && brew link --force llvm; fi
+  - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get update && sudo apt-get install -y llvm-3.5 llvm-3.5-dev; fi
+  - if [ $TRAVIS_OS_NAME == osx ]; then brew update && brew install llvm35 && brew link --force llvm35; fi
```
YorickPeterse: ehm, this won't work once llvm35 is no longer available or when we need LLVM 3.6. I'm pretty sure the "llvm" homebrew package is supposed to be a link to the latest version.

vm/capi/hash.cpp
```diff
@@ -10,6 +10,14 @@ extern "C" {
     return capi_fast_call(rb_cHash, rb_intern("new"), 0);
   }
 
+  VALUE rb_hash_dup(VALUE self) {
+    return capi_fast_call(self, rb_intern("dup"), 0);
+  }
+
+  VALUE rb_hash_freeze(VALUE self) {
+    return capi_fast_call(self, rb_intern("freeze"), 0);
+  }
```
YorickPeterse: These functions need to have specs.

.travis.yml
tamird: llvm35 will be available for ever, it's from homebrew/versions

vm/capi/hash.cpp
tamird: Where should these specs be? The neighbouring functions do not have specs.

tamird:
Removed travis/configure changes, will open a new PR for those. PTAL.

vm/capi/hash.cpp
YorickPeterse: For this you'd use the following files:

* `https://github.com/rubinius/rubinius/blob/master/spec/ruby/optional/capi/hash_spec.rb`: this is used for the actual specs
* `https://github.com/rubinius/rubinius/blob/master/spec/ruby/optional/capi/ext/hash_spec.c`: this is used to expose the C API functions to the above Ruby file
* `https://github.com/rubinius/rubinius/blob/master/spec/ruby/optional/capi/ext/rubyspec.h`: used to "tag" which functions are available

tamird:
Added tests!

vm/capi/hash.cpp
tamird: Thanks for this guidance, done.

tamird:
Green, PTAL.

jemc:
Thanks!

YorickPeterse:
Sweet, thanks!

PR #3433, opened 2015-06-10
Fix OSX build to use llvm35

tamird:
Until RBX supports LLVM 3.6, compiling with homebrew/versions' llvm35 should work painlessly. Also, CI should compile with llvm-3.5 on Linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,5 +50,3 @@ env:
 os:
   - linux
   - osx
-
-osx_image: beta-xcode6.3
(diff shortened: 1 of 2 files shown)


brixen:
Could you revert the change for Linux and concentrate only on OS X in this PR, please.

tamird:
Well, we have a problem. It appears that the test suite has rotted while OSX was disabled in CI. Specifically, the following tests failed on OSX:
```
1) IO#isatty returns true if this stream is a terminal device (TTY) ERROR
2) IO#tty? returns true if this stream is a terminal device (TTY) ERROR
3) Process.setproctitle should set the process title FAILED
4) Global variable $0 actually sets the program name FAILED
```

tamird:
OK, looks like the problem was really with travis' new xcode image; removing that fixed all the things. Green!

PR #3434, opened 2015-06-10
CI with LLVM 3.5 on linux

tamird:


YorickPeterse:
I'm not sure if the gcc fixes are worth the effort. We've been discussing dropping support for gcc in favour of clang for a while now as gcc is a constant source of troubles.

What exactly lead to the LLVM version changes, does installing the regular `llvm` package not work? If this results in LLVM 3.6 being installed that's fine as we're working on upgrading to 3.6 anyway.

brixen:
We don't need to change this right now and as @YorickPeterse mentioned, support for GCC is being removed when LLVM 3.6 support is added.

PR #3439, opened 2015-06-13
Fix Stat::stat_birthtime.

ruipserra:
This commit fixes a build error due to a signature mismatch of the Time::at
method and the way it was being used in Stat::stat_birthtime.

The build error was:
```
vm/builtin/stat.cpp:82:30: error: no viable conversion from 'struct timespec' to 'time64_t' (aka 'long long')
      return Time::at(state, st_.st_birthtimespec);
                             ^~~~~~~~~~~~~~~~~~~~
/Users/ruiserra/code/ruby/ruipserra/rubinius/vm/builtin/time.hpp:40:37: note: passing argument to parameter 'seconds' here
    static Time* at(STATE, time64_t seconds, long nanoseconds = 0);
                                    ^
```

### DIFF
--- a/vm/builtin/stat.cpp
+++ b/vm/builtin/stat.cpp
@@ -79,7 +79,8 @@ namespace rubinius {
 
   Object* Stat::stat_birthtime(STATE) {
     #ifdef HAVE_ST_BIRTHTIME
-      return Time::at(state, st_.st_birthtimespec);
+      struct timespec ts = st_.st_birthtimespec;
+      return Time::at(state, ts.tv_sec, ts.tv_nsec);
     #else
       return Primitives::failure();
     #endif


jemc:
Hm, I wonder why Travis isn't running this to confirm the build.

ruipserra:
Even though 2.2 was added to `.travis.yml` (#3413), I don't see it in the branches list in the UI. And in https://travis-ci.org/rubinius/rubinius/requests it lists this PR as "branch not included or excluded".

jemc:
This looks good to me, although its irksome that Travis won't run it for us.

ruipserra:
Thanks!

brixen:
I've pinged Travis about this (no response), run lint on our .travis.yml, re-read the docs. I have no idea why it's not working. Annoying for sure.

tak1n:
@ruipserra Thanks for fixing this, not sure why I was able to build it with this mismatch

PR #3440, opened 2015-06-13
Delegate Proc#source_location to Method#source_location when the proc was created from a method

ruipserra:


jemc:
I don't see why not...

PR #3448, opened 2015-06-22
Fixed return value for Rubinius::Mirror::Range#step_float_iterations_…

nirvdrum:
…size.

### DIFF
--- a/kernel/common/range_mirror.rb
+++ b/kernel/common/range_mirror.rb
@@ -17,6 +17,8 @@ module Rubinius
         else
           iterations = ((last - first) / step_size + err).floor + 1
         end
+
+        iterations
       end
 
       def step_iterations_size(first, last, step_size)


kernel/common/range_mirror.rb
```diff
@@ -17,6 +17,8 @@ def step_float_iterations_size(first, last, step_size)
         else
           iterations = ((last - first) / step_size + err).floor + 1
         end
+
+        iterations
```
YorickPeterse: Shouldn't the return value of the `if excl` block above already return this?

nirvdrum: If the `if` modifier is false, the result of the expression is `nil`.

nirvdrum: Alternatively, this could be pushed inside the `if` block.  The issue only affects the `excl` case.

Issue #3391, opened 2015-05-07
Define fast-path Symbol#==


jemc:
See https://bugs.ruby-lang.org/issues/11127

YorickPeterse:
Closed by #3449

PR #3449, opened 2015-06-25
Use BasicObject#equal? for Symbol#==.

eregon:
Significantly faster for the case of different symbols:

```ruby
require 'benchmark/ips'
Benchmark.ips do |b|
  b.report("Symbol#==") do
    :a == :b
  end
end
```

Before:
Symbol#==      2.712M (± 3.4%) i/s -     13.589M
After
Symbol#==      6.862M (± 2.6%) i/s -     34.413M

### DIFF
--- a/kernel/common/symbol.rb
+++ b/kernel/common/symbol.rb
@@ -7,6 +7,9 @@ class Symbol
     to_s <=> other.to_s
   end
 
+  # Use equal? for == (and not Comparable version)
+  alias_method :==, :equal?
+
   def capitalize
     to_s.capitalize.to_sym
   end


eregon:
Semantics are not changed, because in Symbol#`<=>` only subclasses of Symbol are allowed, but it is anyway prohibited to make an instance of a subclass.

YorickPeterse:
Yeah, I think I or @jemc bumped into this a while ago which ultimately led to @jemc creating #3391. Since there's no downside to this I don't see why we shouldn't merge this, thanks!

jemc:
Thanks, @eregon - I self-assigned #3391 so that I would get to it when I had time at the end of that week, then completely forgot about it!

PR #3452, opened 2015-06-26
Change Comparable#== to use a recursion guard

eregon:
Changes Comparable#== to use a recursion guard instead of rescuing a SystemStackError.

Stack overflows should generally not be caught and the user #<=> should be called only once (instead of 1214 in my environment).


brixen:
Thanks, @eregon.

PR #3469, opened 2015-07-18
Rename `Float#negative?` to `Float#signbit?`

robin850:
Hello,

Ruby 2.3 ships with a new `Numeric#negative?` method. Rubinius implements this method since b92c84f7, for internal usage though ; it delegates to the `signbit` C++ function.

However, `signbit`'s behavior can't match MRI `Float#negative?`'s one since the former operates on the sign while the latter operates on the value (see [here](http://git.io/vm1fg)).

This makes [a test](https://github.com/rails/rails/blob/0daf1619a0631f7a5022645aca3a39b6d1ab7d07/activesupport/test/core_ext/numeric_ext_test.rb#L444) to fail on the Rails test suite (the [same example](http://en.cppreference.com/w/cpp/numeric/math/signbit#Examples) is present in the `signbit`'s documentation and -0.0 is considered negative this time).

Moreover, if people rely on feature detection rather than version checking, they may wrongly assume that the current running Ruby is 2.3+ checking if a float respond to `#negative?`.

Have a nice day!

### DIFF
--- a/kernel/common/complex.rb
+++ b/kernel/common/complex.rb
@@ -277,7 +277,7 @@ class Complex < Numeric
   def to_s
     result = real.to_s
 
-    if imag.kind_of?(Float) ? !imag.nan? && imag.negative? : imag < 0
+    if imag.kind_of?(Float) ? !imag.nan? && imag.signbit? : imag < 0
       result << "-"
     else
       result << "+"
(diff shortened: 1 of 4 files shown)


YorickPeterse:
Rebased into 533da7eadbf65d7ad14ac74e6d491560c1814293, thanks!

brixen:
@robin850 if people are relying on checking a method for version, please direct them to https://github.com/brixen/redcard. We won't be fixing issues where people wrongly use methods or classes to infer the Ruby version.

PR #3470, opened 2015-07-18
Allow passing strings to #public_constant and #private_constant

robin850:
Hello,

This is a tiny patch that makes `#private_constant` and `#public_constant` accept strings as arguments. The maintained constants table is dealing with symbols but these methods can take strings ; it's pretty common to pass `class.name` as an argument.

MRI is doing the [symbol conversion](https://github.com/ruby/ruby/blob/1782dd8cd753401578ae19ea8be285f9d41bcde5/variable.c#L2593) even though the documentation sets ["symbol" as the given argument](https://github.com/ruby/ruby/blob/1782dd8cd753401578ae19ea8be285f9d41bcde5/variable.c#L2618).

The Active Record test suite doesn't run at all on Rubinius since `has_and_belongs_to_many` relies on this.

Have a nice day.

### DIFF
--- a/kernel/common/module.rb
+++ b/kernel/common/module.rb
@@ -130,6 +130,7 @@ class Module
   end
 
   def private_constant(*names)
+    names = names.map(&:to_sym)
     unknown_constants = names - @constant_table.keys
     if unknown_constants.size > 0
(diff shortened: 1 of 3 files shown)


robin850:
Thanks!

PR #3472, opened 2015-07-21
Implement {Method,UnboundMethod}#super_method

robin850:
Hello,

This is a tiny pull request that adds the `#super_method` method to the `Method` and `UnboundMethod` objects.

With this feature added, the method objects need to keep track of the ancestors chain. This is not a problem when the chain only contain parents of the class. This is, however, when a module is included inside it since we loose the previous ancestry chain.

The method is about [5 times slower](http://git.io/vmhpf) than the MRI implementation though. I don't know if it's acceptable.

Cross-refs #3264.

Have a nice day.

### DIFF
--- a/kernel/common/method.rb
+++ b/kernel/common/method.rb
@@ -158,6 +158,20 @@ class Method
     UnboundMethod.new(@defined_in, @executable, @receiver.class, @name)
   end
 
+  def super_method
+    superclass = @defined_in.direct_superclass
+
+    if superclass
+      mod, entry = superclass.lookup_method(@name)
+
+      if entry && entry.visibility != :undef
+        return Method.new(@receiver, superclass, entry.method, @name)
+      end
+    end
+
+    return nil
+  end
+
 end
 
 ##
@@ -287,5 +301,19 @@ class UnboundMethod
     else
       @defined_in
     end
+  end
+
+  def super_method
+    superclass = @defined_in.direct_superclass
+
+    if superclass
+      mod, entry = superclass.lookup_method(@name)
+
+      if entry && entry.visibility != :undef
+        return UnboundMethod.new(superclass, entry.method, @defined_in, @name)
+      end
+    end
+
+    return nil
   end
 end
--- a/spec/ruby/core/method/super_method_spec.rb
+++ b/spec/ruby/core/method/super_method_spec.rb
@@ -22,4 +22,18 @@ describe "Method#super_method" do
     sss_meth.receiver.should == obj
     sss_meth.name.should == :overridden
   end
+
+  it "returns nil when there's no super method in the parent" do
+    method = Object.new.method(:method)
+    method.super_method.should == nil
+  end
+
+  it "returns nil when the parent's method is removed" do
+    object = MethodSpecs::B.new
+    method = object.method(:overridden)
+
+    MethodSpecs::A.class_eval { undef :overridden }
+
+    method.super_method.should == nil
+  end
 end
--- a/spec/ruby/core/unboundmethod/fixtures/classes.rb
+++ b/spec/ruby/core/unboundmethod/fixtures/classes.rb
@@ -73,7 +73,7 @@ module UnboundMethodSpecs
     def baz(a, b)
       return [__FILE__, self.class]
     end
-    def overriden; end
+    def overridden; end
   end
 
   class B < A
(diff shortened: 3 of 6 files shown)


YorickPeterse:
I can't imagine we don't have anything in place that would make supporting this easier _without_ having to manually track and flush ancestors in Method/UnboundMethod objects. @brixen does anything come to mind?

kernel/common/method.rb
```diff
@@ -158,6 +159,23 @@ def unbind
     UnboundMethod.new(@defined_in, @executable, @receiver.class, @name)
   end
 
+  def super_method
+    @ancestors.each_with_index do |a, i|
```
YorickPeterse: As an aside, please don't use one letter variables. I know the codebase is littered with them but it's a huge pain to constantly try and remember what `a` or `x` or another-one-letter-variable stands for.

kernel/common/method.rb
```diff
@@ -158,6 +159,23 @@ def unbind
...
+      if a.methods.include?(@name)
+        method = a.method(@name).bind(a)
+      elsif a.instance_methods.include?(@name)
+        method = a.instance_method(@name).bind(@receiver)
+      end
+
+      if defined?(method)
+        method.instance_variable_set(:@ancestors, @ancestors.drop(i+1))
```
YorickPeterse: OK if stuff like this is required we'll _definitely_ need to do some extra plumbing to support this nicely. Injecting instance variables into another object is a no-no. Thinking of it this also poses a synchronization problem: what if two threads end up at this code at the same time?

kernel/common/method.rb
```diff
@@ -288,4 +307,21 @@ def owner
...
+      if a.methods.include?(@name)
+        method = a.method(@name)
+      elsif a.instance_methods.include?(@name)
+        method = a.instance_method(@name)
+      end
+
+      if defined?(method)
+        method.instance_variable_set(:@ancestors, @ancestors.drop(i+1))
+        return method
+      end
```
YorickPeterse: Same as my notes above.

YorickPeterse:
Maybe it helps to start with the definition of `super_method`: what is it supposed to do?

jemc:
@robin850 - I agree with @YorickPeterse that it would be best not to use the `@ancestors` variable here.  In fact, I think this could be a much smaller implementation with some of the internal tools Rubinius has available.  Just to give you an idea, one idea that comes to mind would be something like (though I haven't tested this):
```ruby
direct_superclass = @defined_in.direct_superclass
if direct_superclass
  mod, entry = direct_superclass.lookup_method(method_name)
  # Use mod and entry to construct the object to return
end
```

> Maybe it helps to start with the meaning of `super_method`: what is it supposed to do?
There is a spec that this PR is fulfilling (see the removed fail tags).

With that said, the spec looks a bit sparse, so @robin850, if you know more about the behavior of `super_method` under other edge cases please add to the spec.

robin850:
Hey, thank you very much for your quick comments guys! This looks way better with your proposal @jemc, the `@ancestors` thingy was horrible, indeed! The method seems also a bit faster (from ~5 to ~2 times slower than MRI).

> if you know more about the behavior of super_method under other edge cases please add to the spec

At least there were no specs for the case when it returns `nil`. Also, looking at the original commit on the Ruby repository (ruby/ruby@b498159), there were no specs for the case when parent's methods get removed.

This would be cool to test the implementation against a project but It looks like this method is not yet widely used doing a quick search on GitHub ; at least [this test](https://github.com/rails/rails/blob/52cf1a71b393486435fab4386a8663b146608996/activesupport/test/core_ext/marshal_test.rb#L14-L20) passes in the Active Support test suite.

jemc:
Thanks, @robin850 - looks much better to me.  I'll leave it to @YorickPeterse to merge when he also approves.

spec/ruby/core/method/super_method_spec.rb
```diff
@@ -22,4 +22,18 @@
     sss_meth.receiver.should == obj
     sss_meth.name.should == :overridden
   end
+
+  it "returns nil when there's no super method in the parent" do
+    method = Object.new.method(:method)
+    method.super_method.should == nil
+  end
+
+  it "doesn't look for removed methods" do
```
YorickPeterse: I'd change this to `it 'returns nil when the method is removed'` or something along those lines. "doesn't look for..." doesn't really indicate that `super_method` returns in this case.

YorickPeterse:
Other than my little spec description remark this looks fine to me, great work!

robin850:
Fair point! :+1: The patch has been updated, thanks for your quick review guys!

YorickPeterse:
Sweet, thanks!

PR #3476, opened 2015-07-29
Fix Range#bsearch for matching end value in find-minimum mode

### LINKED ISSUE #3458: Range#bsearch range is off by one
`Range#bsearch` ignores the last element:
```
$ ruby -v -e 'p (1..5).bsearch { |i| i >= 5 }'
ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-linux]
5
```
```
$ rbx  -v -e 'p (1..5).bsearch { |i| i >= 5 }'
rubinius 2.5.7 (2.1.0 d4fdeaad 2015-06-17 3.5.2 JI) [x86_64-linux-gnu]
nil
```

Similarly, for `1...5` the last included element `4` is missed.

Also, once this is fixed, maybe `Array#bsearch` should defer to `Range#bsearch` in order to share code?
```ruby
def bsearch
  return to_enum :bsearch unless block_given?
  index = (0...size).bsearch { |i| yield at(i) }
  index ? at(index) : nil
end
```

kachick:
Fixes #3458

### DIFF
--- a/kernel/common/range.rb
+++ b/kernel/common/range.rb
@@ -51,47 +51,47 @@ class Range
 
     last_true = nil
 
-    if max < 0 and min < 0
-      value = min + (max - min) / 2
-    elsif min < -max
-      value = -((-1 - min - max) / 2 + 1)
-    else
-      value = (min + max) / 2
-    end
+    seeker = Proc.new do |current|
+      x = yield current
 
-    while min < max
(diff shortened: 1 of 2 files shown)


kachick:
Thanks!

jc00ke:
Thank you!

PR #3367, opened 2015-03-28
MCJIT Support

YorickPeterse:
This is still very much a work in progress, currently in the early stages.

PR #3479, opened 2015-08-01
Add experimental support for LLVM 3.6

digitalextremist:
Succeeded in builds in my experience.

@brixen would know better than I would the actual consequences of this.

### DIFF
--- a/configure
+++ b/configure
@@ -578,7 +578,7 @@ class Configure
           config = macports_llvm_config
         else
           out = Bundler.with_clean_env { `brew --prefix llvm35` }.chomp
-          config = "#{out}/bin/llvm-config-3.6" if $?.success?
+          config = "#{out}/bin/llvm-config" if $?.success?
         end
       end
     end
--- a/configure
+++ b/configure
@@ -578,7 +578,7 @@ class Configure
           config = macports_llvm_config
         else
           out = Bundler.with_clean_env { `brew --prefix llvm35` }.chomp
-          config = "#{out}/bin/llvm-config-3.5" if $?.success?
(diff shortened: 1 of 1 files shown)


YorickPeterse:
A big pile of changes for LLVM 3.6 can already be found in this pull request: https://github.com/rubinius/rubinius/pull/3367/files

Sadly just changing the version to build with isn't enough as LLVM 3.6 removed the old JIT API that we rely on. I'll close this issue in favour of #3367.

digitalextremist:
@YorickPeterse, damn. I thought this was solvable by magic... but alas, it is not. I reverted to `14.04` of Ubuntu while this is sorted out.

YorickPeterse:
@digitalextremist If available you can install a 3.5 version of LLVM (usually distributions keep a few versions around) and use `./configure --llvm-config=llvm-config3.5` (or something similar) to use an older version of LLVM.

digitalextremist:
It'd be great if the `configure` script could detect older versions. I tried to `rvm install rbx` on a fresh `Ubuntu 15.04 x64` install, and it just choked. That's probably a big problem as people migrate to the next distribution of that Linux distribution.

jc00ke:
Yeah, some detection around `llvm-config` (maybe it would fall back to `llvm-config-3.5` then to `llvm-config-3.4`) would be nice.

larskanis:
This worked for me on Ubuntu-15.04:
``` sh
sudo apt-get install llvm-3.5
rvm install rbx-2.5.8 -C --llvm-config=llvm-config-3.5
```

PR #3489, opened 2015-08-25
Specify `superclass` in respect to `prepend`

benlovell:
This was unspecified/recently merged to ruby/rubyspec. Rubinius does
the Right Thing :tm:

### DIFF
--- a/spec/ruby/core/module/prepend_spec.rb
+++ b/spec/ruby/core/module/prepend_spec.rb
@@ -6,6 +6,10 @@ describe "Module#prepend" do
     Module.should have_public_instance_method(:prepend, true)
   end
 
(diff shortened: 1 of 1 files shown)


YorickPeterse:
For those that didn't do the right thing, what was the behaviour? Would `superclass` return `Module` in that case?

benlovell:
jruby/jruby#3276, yep. For now at least.

YorickPeterse:
Travis failures are due to Homebrew apparently failing to download packages, the rest looks fine. Rebased into dbd33a6dadd407041fcafa6651986ae02551f3ee, thanks!

benlovell:
:heart:

PR #3516, opened 2015-10-10
Remove converting to negative fd

tak1n:
Opening a PR for it because I'm not entirely sure if it should be removed.
Removing it means following [behaviour](http://ruby-doc.org/core-2.2.3/Process.html#method-c-spawn):

Another way to merge multiple file descriptors is [:child, fd]. [:child, fd] means the file descriptor in the child process. This is different from fd. For example, :err=>:out means redirecting child stderr to parent stdout. But :err=>[:child, :out] means redirecting child stderr to child stdout. They differ if stdout is redirected in the child process as follows.

```
# stdout and stderr is redirected to log file.
# The file "log" is opened just once.
pid = spawn(command, :out=>["log", "w"], :err=>[:child, :out])
```


tak1n:
Note:
Travis still doesn't recognize our 2.2 branch.
We should as soon as possible open a pr from 2.2 to master to get 2.2 specs run on travis.

chuckremes:
You should create a spec that shows the failure in Process.spawn. It should go somewhere near here:
https://github.com/rubinius/rubinius/blob/2.2/spec/ruby/shared/process/spawn.rb#L335

Then make your code change and verify the behavior is correct. I don't think this PR will be merged without a spec.

tak1n:
Thx for reviewing accidentally forgot to commit it.

chuckremes:
Looks good to me. Let's leave it for a few days to see if @YorickPeterse or @brixen have comments before merging it.

spec/ruby/shared/process/spawn.rb
```diff
@@ -376,6 +376,14 @@
     @name.should have_data("glark")
   end
 
+  it "redirects STDERR to child STDOUT if :err => [:child, :out]" do
```
YorickPeterse: I would use regular English for this instead, so `if :err is [:child, :out]`, but that's a minor detail.

kernel/common/process_mirror.rb
```diff
@@ -186,7 +186,6 @@ def convert_to_fd(obj, target)
             when 2
               if obj[0] == :child
                 fd = convert_to_fd obj[1], target
```
YorickPeterse: With the check below being removed the assignment can also be removed (as far as I can tell).

YorickPeterse:
Regarding Travis, I can't seem to figure out why it's not building the 2.2 branch. Does running `./bin/mspec ci` pass all specs?

tak1n:
All the specs which were marked as failing (IO popen) and my Process.spawn spec pass.
There are these nil/true/false specs failing, but they were already failing.

YorickPeterse:
:+1: Thanks

PR #3524, opened 2015-11-16
Include zlib linker flag if not building with the vendored one

ahmadsherif:
Apparently building with LLVM enabled includes this flag, probably because of using `--system-libs`, so  building without LLVM fails because it can't link against zlib.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ notifications:
     on_start: always
 
 env:
-  - RVM=2.0.0 LANG="en_US.UTF-8"
+  - RVM=2.2.0 LANG="en_US.UTF-8"
 
 os:
   - linux
--- a/rakelib/blueprint.rb
+++ b/rakelib/blueprint.rb
@@ -179,6 +179,8 @@ Daedalus.blueprint do |i|
     end
     gcc.add_library zlib
     files << zlib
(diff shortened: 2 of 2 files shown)


YorickPeterse:
For some bizarre reason the build keeps failing on OS X when running `gem --version`. Since I don't use OS X I sadly can't test these changes locally, can anybody else using OS X run them to confirm everything is working fine?

YorickPeterse:
cc @brixen @jemc

ahmadsherif:
Looks like Ruby 2.0.0-p576 and RubyGems 2.5.0 are a bad combo on OS X, I think subsequent OS X build will fail for the same reason. Can we bump Ruby version a little?

jemc:
Sorry, no idea about OS X issues - I only build on linux.

YorickPeterse:
@ahmadsherif Bumping it to 2.2 should be fine, you're free to add this to this PR if you'd like (that way we can immediately see if it actually works).

ahmadsherif:
@YorickPeterse All passing except for the clang/linux job, which fails for the same reason as the clang/osx job from the last master build.

YorickPeterse:
@ahmadsherif Restarted build, it's green now. Thanks!

PR #3525, opened 2015-11-18
Reset globally cached klass methods in Module#prepend_features

ahmadsherif:


jemc:
@ahmadsherif - Can you take a moment to explain why this is needed when the cache is also cleared [from `include_modules_from`](https://github.com/rubinius/rubinius/blob/4eec62a07d6b3cbdfee3c4ce0b6e6a18498641a4/kernel/common/type.rb#L333), which is called by `prepend_features`?

ahmadsherif:
@jemc Sure. `include_modules_from` only clears the cache of the newly prepended methods, other cached methods from `klass` are untouched, which could result in a wrong lookup result (like the example in the [commit message](https://github.com/ahmadsherif/rubinius/commit/5c4b45f0aa1793b31bb9ce6d1bf3e5ef631a7120)).

YorickPeterse:
Thanks!

PR #3534, opened 2015-12-13
Compile fails on arm because of unknown opcode.

dafoxia:
"stmeqia" is an mistyped and unknown opcode.
Replaced with "stmiaeq".
Compiling ffi on ARM now possible.
Maybe an update do a newer version of libffi is also possible.

Compiling of rubinius failed later with "no memory barrier implementation" in atomic.hpp
:(

### DIFF
--- a/vendor/libffi/src/arm/sysv.S
+++ b/vendor/libffi/src/arm/sysv.S
@@ -396,7 +396,7 @@ LSYM(Lbase_args):
 	beq	LSYM(Lepilogue_vfp)
 
 	cmp	r3, #FFI_TYPE_SINT64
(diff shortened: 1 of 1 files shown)


YorickPeterse:
Is it known if these changes are also available in the upstream FFI, or if they can be submitted to the upstream?

dafoxia:
On the archive download link (https://sourceware.org/libffi/) libffi-3.2.1 the mistake is still in.
In actual git-repo this file is almost completly different and the hard-float-section seems removed.

YorickPeterse:
Rebased into baf794d0f6c23f029872575cccb9bd2b113723c4, thanks!

PR #3541, opened 2015-12-25
Update raising of WaitReadable/WaitWritable exceptions

YorickPeterse:
This changes `IO#read_nonblock` and `IO#write_nonblock` to raise the correct exceptions (using somewhat more flexible code) as well as defining extra `WaitReadable` / `WaitWritable` exception classes to be used by rubysl-socket.

### DIFF
--- a/vm/builtin/io.cpp
+++ b/vm/builtin/io.cpp
@@ -727,7 +727,7 @@ namespace rubinius {
     int res = ::select(fd+1, &set, 0, 0, &tv);
 
     if(res == 0) {
-      Exception::errno_eagain_error(state, "no data ready");
+      Exception::errno_wait_readable(state, EAGAIN);
       return 0;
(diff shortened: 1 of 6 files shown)


vm/builtin/io.cpp
```diff
@@ -727,7 +727,7 @@ namespace rubinius {
     int res = ::select(fd+1, &set, 0, 0, &tv);
 
     if(res == 0) {
-      Exception::errno_eagain_error(state, "no data ready");
+      Exception::errno_wait_readable(state, EAGAIN);
```
YorickPeterse: This is a bit of a hack because `errno` is not set when `select` returns 0. Without explicitly passing `EAGAIN` we'd have to handle this case in `Exception::errno_wait_readable` which I think is the wrong place to do so (since this behaviour is very specific to `select`).

PR #3543, opened 2015-12-26
Allow #instance_eval to evaluate procs originated from methods

ahmadsherif:


spec/ruby/core/kernel/instance_eval_spec.rb
```diff
@@ -138,4 +138,13 @@ class B; end
       (1 << 64).instance_eval { def foo; end }
     end.should raise_error(TypeError)
   end
+
+  it "evaluates procs originated from methods" do
```
YorickPeterse: This should be "originating from methods".

ahmadsherif: Done.

YorickPeterse:
Rebased into f0d20890786123bd4870cc9871ce0a8ebbfea16f and a246ed6e32f97cd70ca703f814e4164b81f5871b, thanks!

PR #3548, opened 2016-01-03
Break up old/new license & clarify what we use

YorickPeterse:
I decided to move the BSD license to a separate file as I feel uncomfortable adding any extra clauses to the MPL 2.0 license. I also feel it would be too easy to miss the BSD license at the bottom of the already long MPL 2.0 license. As far as I'm aware the BSD license does not disallow renaming license files, thus I believe we should be fine.

cc @brixen @hintjens

### DIFF
--- a/BSD_LICENSE
+++ b/BSD_LICENSE
@@ -1,0 +1,27 @@
+BSD 3-clause License
+
+Copyright (c) 2007-2015, Evan Phoenix and contributors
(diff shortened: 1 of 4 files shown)


brixen:
Thanks @YorickPeterse, that looks more clear.

jemc:
Thanks, @YorickPeterse!

PR #3552, opened 2016-01-06
If on OS X, check for homebrew-supplied openssl libraries

andre-richter:


YorickPeterse:
@brixen Leaving this up to you since I _think_ you were discussing something about this. I'm also not a Mac user so I'm not really sure what the context and the likes behind this is.

andre-richter:
Hmm, travis timed out when building a native extension. Doesn't look like it has anything to do with my patch, though.

YorickPeterse:
@andre-richter I restarted the build.

brixen:
@andre-richter thanks for working on this. There's another place we check `@darwin` for the LLVM stuff. It's not critical that you do this, but could you take a little time to standardize both those checks? They are similar and having everything a bit more standard (ie consistent) will help when I port this to Bash soon.

brixen:
@andre-richter thanks!

Issue #3554, opened 2016-01-08
Issue with Enumerable


ioquatix:
When using `to_enum`, yield with multiple args is not handled correctly?

```ruby
class Bob
	def foo(*args)
		return to_enum(:foo, *args) unless block_given?
		
		paths = []
		
		3.times do
			yield args.shift, args.shift
		end
	end
end

bob = Bob.new
things = bob.foo(1, 2, 3, 4, 5, 6)

things.each do |thing|
	puts thing.inspect
	# Correct behavior:
	# MRI: gives 1, 3, 5
	# RBX: gives 1, 3, 5
end

puts things.first.inspect
# Odd behavior:
# MRI: [1, 2]
# RBX: 1
```

ioquatix:
After reviewing this issue I found that the problem is with `Enumerable::next`. It doesn't work correctly for the base case. This monkey patch fixes the issue:

```ruby
module Enumerable
	def first(n=1)
		return __take__(n) unless n == 1
		each { |*obj| return *obj }
		nil
	end
end
```

ioquatix:
So, basically the problem is to do with `|*obj|`. I took a look in the code and it might affect other functions too, e.g. `cycle`.

brixen:
@ioquatix thanks for reporting this. Do you know if there is a spec that describes this behavior?

ioquatix:
I have no idea where to look for an official spec but I assume that the behaviour should match MRI? The above code is a basic outline of what a spec might look like, but as I said there are other methods, e.g. `cycle` that are affected.

ioquatix:
Okay, here is an example of the issue with count:

```ruby
#!/usr/bin/env ruby

class Bob
    def foo(*args)
        return to_enum(:foo, *args) unless block_given?

        paths = []

        3.times do
            yield args.shift, args.shift
        end
    end
end

bob = Bob.new
things = bob.foo(1, 2, 3, 4, 5, 6)

puts things.count([1, 2])
# MRI: gives 1
# RBX: gives 0
```

ioquatix:
Here is an example showing `slice_before` is broken:

```ruby
#!/usr/bin/env ruby

class Bob
    def foo(*args)
        return to_enum(:foo, *args) unless block_given?

        paths = []

        3.times do
            yield args.shift, args.shift
        end
    end
end

bob = Bob.new
things = bob.foo(1, 2, 3, 4, 5, 6)

more = things.slice_before do |a, b|
	a == 3 and b == 4
[...]

PR #3556, opened 2016-01-08
Fix Enumerable#first block arities

kachick:
I guess this fixes https://github.com/rubinius/rubinius/issues/3554

### DIFF
--- a/kernel/common/enumerable.rb
+++ b/kernel/common/enumerable.rb
@@ -563,7 +563,9 @@ module Enumerable
 
   def first(n=undefined)
     return __take__(n) unless undefined.equal?(n)
-    each { |element| return element }
+    each do
+      return Rubinius.single_block_arg
+    end
     nil
   end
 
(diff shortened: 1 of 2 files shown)


ioquatix:
Awesome, thanks so much. I guess we should try to evaluate #cycle or any other similar methods (#count?)

YorickPeterse:
Rebased into https://github.com/rubinius/rubinius/compare/87ef3834e9202120a5da088b50a03b04bd304e3e...86bf3170ea8f75911616535ee2659f9666e6b14b, thanks!

> I guess we should try to evaluate #cycle or any other similar methods (#count?)

Yes, but those should go into a separate PR.

ioquatix:
Sure they would be a separate PR, but the Issue which is now closed also raised those points. Should I create a new issue?

YorickPeterse:
@ioquatix If other methods suffer from the same problem then yes, an issue would be appreciated.

Issue #3456, opened 2015-07-07
Rubinius behaves differently than Ruby wrt. use of to_ary/to_s


wkoszek:
I pushed to Git my Ruby exercises I wrote a while back when learning Ruby.

git clone git://github.com/wkoszek/ruby_edu.git

I went and enabled Travis-CI builds for both Ruby 2.1 and Rubinius. 

https://travis-ci.org/wkoszek/ruby_edu/jobs/69929229
https://travis-ci.org/wkoszek/ruby_edu/jobs/69929231

Ruby passes, while Rubinius is failing. Looks like to_s function is called where to_ary is expected, assuming Ruby 2.1 is a reference result.

Upon cloning, do "make". It will run Ruby on each .rb file and generate .o file, which is a result. Then the .o file is compared against corresponding .t file (expected result).

wkoszek:
Is this problem also related with Travis using out-of-date Rubinius?

(more details on outdated Travis: https://github.com/rubinius/rubinius/issues/3457)

brixen:
This build is using 2.5.7 https://travis-ci.org/wkoszek/ruby_edu/jobs/69929231#L104 so it does not look related to the version problem in #3457.

andre-richter:
Interesting find, thank you!
Has been fixed in https://github.com/rubinius/rubinius/commit/648a80cbd52ca25495680f5eefc469e024b106e1.

PR #3558, opened 2016-01-09
puts: Check if argument responds to :to_ary, instead of checking its type being Array

andre-richter:
Fixes interesting find in #3456


brixen:
Thanks!

PR #3564, opened 2016-01-13
Implemented Numeric#positive? and Numeric#negative?

Locke23rus:


jc00ke:
There are a mix of styles in the specs...

```ruby
x.negative?.should == false
# and
x.negative?.should be_false
```

We should use just one style.

spec/ruby/core/bignum/negative_spec.rb
```diff
@@ -0,0 +1,9 @@
+require File.expand_path('../../../spec_helper', __FILE__)
+
+describe "Bignum#negative?" do
+  it "returns true if self is less than 0" do
```
jc00ke: The `x.negative?.should == false` expectations don't exercise what the spec description describes. :confused:

Locke23rus:
@jc00ke which style should I use `== true/false`?

spec/ruby/core/float/negative_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      +1.0,
+      +0.0,
+      -0.0,
+      +(0.0.next_float),
+      +Float::INFINITY,
+      Float::NAN
+    ]
+
+    not_negative_numbers.each do |number|
+      number.negative?.should == false
```
jc00ke: This expectation doesn't exercise the description either.

jc00ke:
@Locke23rus I see `be_[true|false]` is used in `Integer` specs (`Integer#even?` for instance) but I see `should == false` used in other places. I'm partial to `be_[true|false]` personally.

Locke23rus:
@jc00ke thanks, request is updated.

brixen:
Thanks!

PR #3565, opened 2016-01-13
Make rb_rescue returns the break value when passed function yields to a block with a break

ahmadsherif:


vm/capi/kernel.cpp
```diff
@@ -60,8 +60,7 @@ extern "C" {
       ep.pop(env);
 
       if(env->state()->thread_state()->raise_reason() != cException) {
```
andre-richter: I guess it would be safer to check for `cBreak` explicitly and only then return the break value instead of doing it for all kinds of exceptions?

I think a switch() is reasonable here.

ahmadsherif: I guess we're supposed to return the `break` value here, as there's a handling for it (and other raise reasons) [in the interpreter loop](https://github.com/rubinius/rubinius/blob/92ad6b4a69977f4814d8cba96301cc08b1ec417b/vm/instructions.cpp#L147), which is what I'm trying to do here (making execution reach that point).

andre-richter: Ah, in this case ignore my comment, I misunderstood :)

brixen:
@ahmadsherif nice catch! Thanks!

PR #3567, opened 2016-01-14
Set thread's group in rb_thread_create

ahmadsherif:


spec/ruby/optional/capi/thread_spec.rb
```diff
@@ -84,5 +84,9 @@ def call_capi_rb_thread_wakeup
 
       lambda { thr.join }.should raise_error(NotImplementedError)
     end
+
+    it "sets the thread's group" do
+      @t.rb_thread_create(lambda {}, nil).group.should_not be_nil
```
YorickPeterse: Is it possible here to test for the type that it _does_ return, instead of merely testing if it's not nil?

ahmadsherif: Sure, done.

spec/ruby/optional/capi/thread_spec.rb
```diff
@@ -84,5 +84,10 @@ def call_capi_rb_thread_wakeup
 
       lambda { thr.join }.should raise_error(NotImplementedError)
     end
+
+    it "sets the thread's group" do
+      thread_group = @t.rb_thread_create(lambda {}, nil).group
+      thread_group.class.should == ThreadGroup
```
YorickPeterse: Being picky here, but you can use `.....group.should be_an_instance_of(ThreadGroup)` if I'm not mistaken.

ahmadsherif: No problem, done.

YorickPeterse:
Thanks!

PR #3568, opened 2016-01-14
Implement Hash#to_proc

ahmadsherif:


kernel/common/hash.rb
```diff
@@ -632,6 +632,9 @@ def [](key)
     end
   end
 
+  # Used internally to get around subclasses redefining #[]
+  alias_method :__get__, :[]
+
```
YorickPeterse: It's better to either use a mirror (and have `#[]` also use that), or simply make the statement that if people overwrite `#[]` in a subclass its their responsibility to also ensure `#to_proc` works. I'd prefer the latter since the former is an endless game of whack-a-mole we can't win.

ahmadsherif: Assuming I went the latter, where should I make such statement? A comment on `to_proc` perhaps?

YorickPeterse: There's no need to document this anywhere in particular as the statement applies to the Ruby code as a whole (including any existing code). We'll simply inform people about this whenever they run into any problems.

ahmadsherif: Alright, done.

YorickPeterse:
Thanks!

PR #3575, opened 2016-01-15
Implement Enumerable#grep_v

ahmadsherif:


kernel/common/enumerable.rb
```diff
@@ -314,6 +314,28 @@ def grep(pattern)
     ary
   end
 
+  def grep_v(pattern)
+    ary = []
+
+    if block_given?
```
YorickPeterse: The overhead of `block_given?` isn't that dramatic so I think this can just be changed to something like:

```ruby
if block_given?
  ary << yield element
else
  ary << element
end
```

spec/ruby/core/enumerable/grep_v_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      def ===(obj)
+        obj == '2'
+      end
+    end
+    enum = EnumerableSpecs::Numerous.new('2', 'a', 'nil', '3', false)
+
+    enum.grep_v(pattern_cls.new).should == ['a', 'nil', '3', false]
+  end
+
+  it "returns an array of all elements not === pattern passed through the block" do
```
YorickPeterse: This description is a little bit confusing. I think something like `of all elements not equal to the pattern ....` would be better (at least if that's what the description is meant to state).

kernel/common/enumerable.rb
ahmadsherif: I kinda followed `grep` implementation, is it OK if I changed it there too in this PR?

spec/ruby/core/enumerable/grep_v_spec.rb
ahmadsherif: Wouldn't using "equal" cause confusion too (i.e. "==" is referred to as equal)?

kernel/common/enumerable.rb
YorickPeterse: Sure.

spec/ruby/core/enumerable/grep_v_spec.rb
YorickPeterse: Both `==` and `===` test for equality, `===` mostly exists so subclasses of `Object` can overwrite it. As such I think just using "equal" in the description should be good enough.

ahmadsherif: Done.

kernel/common/enumerable.rb
ahmadsherif: Done.

YorickPeterse:
Thanks!

PR #3576, opened 2016-01-16
Dig with no arguments should raise ArgumentError

ndnenkov:
Trying to call `#dig` methods using cruby 2.3.0 + gives you:
```ruby
ArgumentError: wrong number of arguments (given 0, expected 1+)
```

However, the current Rubinius implementation will instead try to use `nil` as a key. This is inconsistent and unintuitive:
```ruby
[].dig # => TypeError: Coercion error: nil.to_int => Fixnum failed
{}.dig # => nil
{nil => 42}.dig # => 42
```

### DIFF
--- a/kernel/common/array.rb
+++ b/kernel/common/array.rb
@@ -597,13 +597,13 @@ class Array
     self
   end
 
-  def dig(*sequence)
-    item = self[sequence.shift]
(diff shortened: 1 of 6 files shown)


YorickPeterse:
Rebased into master, thanks!

PR #3577, opened 2016-01-16
Implement Enumerator::Lazy#grep_v

ahmadsherif:


spec/ruby/core/enumerator/lazy/grep_v_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    ret = @yieldsmixed.grep_v(Object) {}
+    ret.should be_an_instance_of(enumerator_class::Lazy)
+    ret.should_not equal(@yieldsmixed)
+
+    ret = @yieldsmixed.grep_v(Object)
+    ret.should be_an_instance_of(enumerator_class::Lazy)
+    ret.should_not equal(@yieldsmixed)
+  end
+
+  it "sets nil to size" do
```
brixen: I think this should be `it "sets #size to nil"` if I'm reading it right. @ahmadsherif I'm adding you Rubinius collaborators. I'll merge this. You can push an update to change the spec description.

brixen:
Thanks!

PR #3578, opened 2016-01-16
Implement File.mkfifo

ahmadsherif:


kernel/common/file.rb
```diff
@@ -1195,6 +1195,21 @@ def self.sticky?(file_name)
     return false
   end
 
+  def self.mkfifo(file_name, mode = 0666)
+    raise NotImplementedError, "mkfifo is not implemented on this platform" unless Rubinius::HAVE_MKFIFO
```
YorickPeterse: I think it's better to only define this method when `mkfifo` is actually present, opposed to checking for it on every call.

ahmadsherif: Won't this behavior be incompatible with MRI?

brixen: MRI has a special C function that is put in the method table when the method is not implemented on that platform. We do this for FFI, but we don't have a generic way to do it yet.

For now, this is fine. We already do it in a few places and if someone is doing mkfifo in a tight loop, they probably need to refactor something.

brixen:
Thanks!

PR #3579, opened 2016-01-20
Add chunk_while

jsyeo:


kernel/common/enumerable.rb
```diff
@@ -212,6 +212,12 @@ def slice_when(&block)
     end
   end
 
+  def chunk_while(&block)
+    raise ArgumentError, "wrong number of arguments (0 for 1)" unless block_given?
```
jemc: This doesn't strike me as a particularly accurate or helpful error message.

Even if it's the same as the one given by MRI, I'd prefer to see it made clear that a *block is required*.  The current error message makes it look like a regular argument is required.

brixen: Hm, interesting that this doesn't return an Enumerator when no block is given. I see this in the 2.3.0 `ri chunk_while` docs:

```
= .chunk_while

(from ruby core)
=== Implementation from Enumerable
------------------------------------------------------------------------------
  enum.chunk_while {|elt_before, elt_after| bool } -> an_enumerator

------------------------------------------------------------------------------
```

Assuming this should raise an ArgumentError, I agree with @jemc here, we should use a more helpful error message.

brixen: Ah, I see now why the block is required. So yeah, we should fix the error message.

jsyeo: What should the error be? How about "Block required"?

jsyeo: Or how about "no block given" for consistency with `chunk` in https://github.com/rubinius/rubinius/blob/master/kernel/common/enumerable.rb#L8

jsyeo:
Changed error message to "no block given".

kernel/common/enumerable.rb
jemc: :+1:

kernel/common/enumerable.rb
```diff
@@ -212,6 +212,12 @@ def slice_when(&block)
     end
   end
 
+  def chunk_while(&block)
+    raise ArgumentError, "no block given" unless block_given?
+
+    slice_when { |before, after| !(block.call(before, after)) }
```
YorickPeterse: The explicit block isn't needed here, this can just be:

    !(yield before, after)

brixen:
Thanks!

PR #3580, opened 2016-01-21
Remove initial state from slice_before and chunk

jsyeo:
For [Feature #10958](https://bugs.ruby-lang.org/issues/10958)

### DIFF
--- a/kernel/common/enumerable.rb
+++ b/kernel/common/enumerable.rb
@@ -136,18 +136,18 @@ module Enumerable
     h
   end
 
-  def slice_before(arg = undefined, &block)
-    if block_given?
-      has_init = !(undefined.equal? arg)
-    else
-      raise ArgumentError, "wrong number of arguments (0 for 1)" if undefined.equal? arg
-      block = Proc.new{ |elem| arg === elem }
-    end
+  def slice_before(pattern = undefined, &block)
+    pattern_given = !(undefined.equal? pattern)
+
+    raise ArgumentError, "cannot pass both pattern and block" if pattern_given && block_given?
+    raise ArgumentError, "wrong number of arguments (0 for 1)" if !pattern_given && !block_given?
+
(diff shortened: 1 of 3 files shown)


kernel/common/enumerable.rb
```diff
@@ -136,18 +130,18 @@ def group_by
...
-      has_init = !(undefined.equal? arg)
-    else
-      raise ArgumentError, "wrong number of arguments (0 for 1)" if undefined.equal? arg
-      block = Proc.new{ |elem| arg === elem }
-    end
+  def slice_before(pattern = undefined, &block)
+    pattern_given = !(undefined.equal? pattern)
+
+    raise ArgumentError, "cannot pass both pattern and block" if pattern_given && block_given?
+    raise ArgumentError, "wrong number of arguments (0 for 1)" if !pattern_given && !block_given?
```
brixen: @jsyeo what about `"a pattern or a block must be provided"` for this to explain the error better?

Also, I've added you to the Rubinius contributors. Feel free to merge this and new commits if you feel confident in them. If you'd like feedback or are unsure, please use a PR.

Thanks!

jsyeo: > what about "a pattern or a block must be provided" for this to explain the error better?

Sounds good! Let me use that.

>  Feel free to merge this and new commits if you feel confident in them. If you'd like feedback or are unsure, please use a PR.

Oh okay, got it!

jemc:
Isn't this an example of a breaking change?

How do we handle deprecation in this case, as per [the versioning rules](http://rubinius.com/2015/09/22/major-minor-maximize-delivering-features-minimize-trouble/)?

YorickPeterse:
@jemc It seems the initial state argument has been removed as of Ruby 2.3, while it was deprecated starting with 2.2. According to our versioning rules we should add a deprecation first and _then_ remove it. Having said that, since we aren't going to "deploy" all Ruby 2.3 support in a single release I'm not sure how useful this would be.

Issue #3559, opened 2016-01-09
Need Ruby 2.3 specs


chuckremes:
Pulled from Ruby [News](https://github.com/ruby/ruby/blob/trunk/doc/NEWS-2.3.0)

### Language changes

* frozen-string-literal pragma:

  - [x] ~~new pragma, frozen-string-literal has been experimentally introduced.
    [Feature #8976](https://bugs.ruby-lang.org/issues/8976)~~
  - [x] ~~besides, --enable/--disable=frozen-string-literal options also have
    been introduced. [Feature #8976](https://bugs.ruby-lang.org/issues/8976)~~
  - [x] ~~command line options --debug or --debug=frozen-string-literal enable
    additional debugging mode which shows created location with at frozen
    object error (RuntimeError).
    [Feature #11725](https://bugs.ruby-lang.org/issues/11725)~~

We're not going to port over MRI's weird String hacks, we can either make the compiler smarter (removing the need for developers to add anything) or have the JIT optimize code where needed.

* safe navigation operator:

  - [x] new method call syntax, `object&.foo', method #foo is called on
    `object' if it is not nil.
    this is similar to `try!' in Active Support, except:
    - [x] method name is syntactically required
          obj.try! {} # valid
          obj&. {}    # syntax error
    - [x] arguments are evaluated only if a call is made:
          obj.try!(:foo, bar())  # bar() is always evaluated
          obj&.foo(bar())        # bar() is conditionally evaluated
    - [x] attribute assignment is valid
          obj&.attr += 1
    [Feature #11537](https://bugs.ruby-lang.org/issues/11537) Moved to https://github.com/rubinius/rubinius/issues/3601

* the did_you_mean gem:

  - [x] ~~When a NameError or NoMethodError occurs because of a typo in the name,
    the did_you_mean gem automatically suggests other names similar to the
    method name.~~ People can just install this Gem themselves

      "Yuki".starts_with?("Y")
      # => NoMethodError: undefined method `starts_with?' for "Yuki":String
      #    Did you mean?  start_with?

* indented here document:

  - [x] new string literal, here document starts with `<<~`.
    refer doc/syntax/literals.rdoc for more details.
    [Feature #9098](https://bugs.ruby-lang.org/issues/9098) Moved to https://github.com/rubinius/rubinius/issues/3602

### Core classes updates (outstanding ones only)

* ARGF

  - [x] ARGF.read_nonblock supports `exception: false' like IO#read_nonblock.
[...]

PR #3583, opened 2016-01-21
Implement String.new encoding optional argument

tak1n:
Related to #3559

### DIFF
--- a/kernel/common/string.rb
+++ b/kernel/common/string.rb
@@ -42,8 +42,9 @@ class String
     alias_method :dup, :clone
   end
 
-  def initialize(arg = undefined)
+  def initialize(arg = undefined, encoding: nil)
     replace arg unless undefined.equal?(arg)
+    self.force_encoding(encoding) if encoding
     self
   end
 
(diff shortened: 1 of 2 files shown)


kernel/common/string.rb
```diff
@@ -42,8 +42,9 @@ def clone
     alias_method :dup, :clone
   end
 
-  def initialize(arg = undefined)
+  def initialize(arg = undefined, encoding: nil)
     replace arg unless undefined.equal?(arg)
+    self.force_encoding(encoding) if encoding
```
YorickPeterse: What's the reason for this using `force_encoding` and not `encode!`?

tak1n: There is no reason for that I was not aware of the `encode!` method.
Should I change it to use `encode!` ?

tak1n: After digging into mri's c code @YorickPeterse and me came to the conclusion to use force_encoding here.
MRI is using `rb_enc_associate`: https://github.com/ruby/ruby/blob/trunk/string.c#L1354
Which is also used in `force_encoding`: https://github.com/ruby/ruby/blob/trunk/string.c#L8671

PR #3584, opened 2016-01-22
Fix issue in slice_(after|before) when iterator method yields more than one value

jsyeo:
So I took some time to look at the issue that was reported by @ioquatix in https://github.com/rubinius/rubinius/issues/3554#issuecomment-170545804 and I found out that it affects all our slice_* methods. Unfortunately, I didn't spend much time to figure out how to fix this but I will come back to this when I have the time. As for this PR, I hope someone can give me some feedback with regard to the clarity of the description of the specs and the readability of the specs. I am not sure if the spec is readable enough in this current state.

### DIFF
--- a/spec/ruby/core/enumerable/slice_after_spec.rb
+++ b/spec/ruby/core/enumerable/slice_after_spec.rb
@@ -46,6 +46,15 @@ describe "Enumerable#slice_after" do
     end
   end
 
+  describe "when iterator method yields more than one value" do
+    it "should process all yielded values" do
+      def foo
+        yield 1, 2
+      end
+      to_enum(:foo).slice_after { true }.to_a.should == [[[1, 2]]]
+    end
+  end
+
   it "raises an ArgumentError when given an incorrect number of arguments" do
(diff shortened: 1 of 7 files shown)


spec/ruby/core/enumerable/slice_after_spec.rb
```diff
@@ -46,6 +46,15 @@
     end
   end
 
+  describe "when iterator method yields more than one value" do
```
YorickPeterse: "when an iterator method" reads more pleasant.

spec/ruby/core/enumerable/slice_after_spec.rb
```diff
@@ -46,6 +46,15 @@
...
+    it "should process all yielded values" do
```
YorickPeterse: `it "processes ...`. I've always disliked the idea of using `it "should ...` as it suggests it may be OK for the code to do otherwise.

brixen: Yes, we don't use `it "should..."` in specs.

PR #3589, opened 2016-01-25
Add keyword arguments support for zsuper

ahmadsherif:
I'm not sure if the direction I took for solving this problem (using the placeholder, etc.) is right, so a feedback would be appreciated. ~~Also, I didn't remove the "fails" tags for specs related to this problem because the changes [here](https://github.com/rubinius/rubinius-ast/pull/7) are needed for the specs to pass.~~

### DIFF
--- a/spec/jit/method_spec.rb
+++ b/spec/jit/method_spec.rb
@@ -155,5 +155,29 @@ with_feature :jit do
         end
       end
     end
+
+    context "to m(a:, b: 'b', **) super end" do
+      before :each do
+        parent = Class.new do
+          def m(**args) args end
+        end
+
+        child = Class.new(parent) do
+          def m(a:, b: 'b', **) super end
+        end
+
+        @o = child.new
+
+        jit(@o, :m) { @o.m(a: "a", c: "c") }
+      end
+
+      it "compiles" do
+        @o.method(:m).executable.jitted?.should be_true
+      end
+
+      it "returns all keyword arguments passed including default ones" do
+        @o.m(a: "a", c: "c").should == {a: "a", b: "b", c: "c"}
+      end
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


YorickPeterse:
@brixen I fiddled around with this last week but wasn't really sure about our current setup to begin with. For example, when keyword arguments are present we seem to allocate a Hash (basically the old `options = {}` argument) instead of using proper keyword arguments.

Since I'm not really sure about things I'll leave this one up to you.

brixen:
@ahmadsherif I'll pull the AST stuff into https://github.com/rubinius/rubinius-code ASAP and we can work from there.

@YorickPeterse there's no such thing as proper keyword arguments, it's janky Hash all the way down in Ruby. We may be able to do some optimizations once we start cleaning up the instruction set.

YorickPeterse:
With the stuff pulled in, can this be merged?

brixen:
@YorickPeterse the changes to the functions need to be represented in the JIT as well. A good first step is writing JIT specs for these cases.

brixen:
s/functions/instructions/

ahmadsherif:
@brixen Apparently [JIT is disabled](https://github.com/rubinius/rubinius/commit/ae81a26ec0b3c81aa287db053b4d11e634aa67e4) for methods with keyword arguments, should we fix this first? I would like to try but I don't know what's "compile policy checks" :).

brixen:
@ahmadsherif sorry, somehow I didn't see your last message. I merged this since the JIT is disabled right now and we'll fix this up as part of re-enabling the JIT. Also, it's a good time to look at integrating keywords into the Arguments structure so eg we can optimize sites that have explicit keywords with the inline cache mechanism to pass the arguments straight through when they are not captured in a kwrest argument.

PR #3591, opened 2016-01-28
Fix count when enumerable yields multiple values

jsyeo:


spec/ruby/core/enumerable/count_spec.rb
```diff
@@ -43,8 +43,15 @@ def size
     @numerous.count(4){|x| x%2==0 }.should == 1
   end
 
-  it "gathers initial args as elements when each yields multiple" do
-    multi = EnumerableSpecs::YieldsMulti.new
-    multi.count {|e| e == 1 }.should == 1
+  context "when each yields multiple" do
```
YorickPeterse: We generally use `describe` instead of `context`.

jsyeo:
This will fix https://github.com/rubinius/rubinius/issues/3554#issuecomment-170544407

ioquatix:
Nice work guys :)

Issue #3592, opened 2016-01-29
Possible Enumerable#sort_by bug


asppsa:
Hello,

I think I may have found an issue with [Enumerable#sort_by][1].  When a block is given, this method calls `#map` to create another enumerable of objects to sort. It then calls `#sort!` on that new enumerable.  The problem is that `#sort!` is only defined on `Array`, so in cases where `#map` does not return an array, we get a `NoMethodError`.  Simple example (in rbx-3.5.0):

~~~
irb(main):001:0>  [1,2,3,4,5].lazy.sort_by(&:itself)
NoMethodError: undefined method `sort!' on an instance of Enumerable::Enumerator::Lazy.
        from kernel/delta/kernel.rb:78:in `sort! (method_missing)'
        from kernel/common/enumerable.rb:345:in `sort_by'
        from (irb):1
~~~

`#sort_by` with a block is always supposed to return an `Array`, so I believe a simple `#to_a` call after the `#map` will solve this.  I will have a go at fixing this now ...

[1]: https://github.com/rubinius/rubinius/blob/master/kernel/common/enumerable.rb#L351

YorickPeterse:
Fixed by #3593 correct?

asppsa:
Yep!

PR #3593, opened 2016-01-29
Fix Enumerable#sort_by when #map does not return an array

asppsa:
See #3592 for a description of the issue fixed here.


brixen:
Thanks!

PR #3595, opened 2016-01-29
Added ARGF.read_nonblock

YorickPeterse:
@ahmadsherif could you check if any specs from your Gist (https://gist.github.com/ahmadsherif/998858e65899716edd7d) and add them to this merge request if they're useful enough? I noticed you also moved a bunch of things around, I didn't include those because they don't seem to be related to read_nonblock itself.

### DIFF
--- a/kernel/common/argf.rb
+++ b/kernel/common/argf.rb
@@ -363,6 +363,29 @@ module Rubinius
       output
     end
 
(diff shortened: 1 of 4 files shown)


ahmadsherif:
@YorickPeterse I don't see more specs to add from the gist :). The stuff that was moved was a common behavior between `readpartial` and `read_nonblock`.

PR #3617, opened 2016-02-05
Clean up define method

tak1n:
First I tried to ship most of the logic towards executables and use `Method#for_define_method` and `UnbindMethod#for_define_method` as thin wrapper around `Executable#for_define_method`.
The result can be seen here: https://github.com/rubinius/rubinius/commit/7cd43157d6efe0a4b4a534ef944df8563a062979

When the need to support Proc again came up it was clear that shipping most of the stuff into executables was not quite doable, or at least I was not able to do it properly.
Therefore I moved it up into `Method` and `UnboundMethod` where again we have nested conditionals (but at least no more nested conditionals in a big fat case statement)

As for `BlockEnvironment::AsMethod#scope` I'm not sure if this should return nil or the scope of `@block_env`. In the previous implementation `define_method` for a `BlockEnvironment::AsMethod` executable used nil as scope. Also when changing it to `@block_env.scope` all ci specs are still passing so probably this means there are missing specs for that.

### DIFF
--- a/core/block_environment.rb
+++ b/core/block_environment.rb
@@ -104,7 +104,7 @@ module Rubinius
       end
 
       def scope
-        nil
+        @block_env.scope
       end
     end
 
--- a/core/block_environment.rb
+++ b/core/block_environment.rb
@@ -103,10 +103,8 @@ module Rubinius
         @block_env.line
       end
 
-      def for_define_method(name, meth)
-        code  = Rubinius::DelegatedMethod.new(name, :call_on_instance, meth, true)
-
-        [code, nil]
+      def scope
+        nil
       end
     end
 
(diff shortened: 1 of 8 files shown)


tak1n:
Regarding to https://github.com/rubinius/rubinius/commit/d59ca90e9841253235331ae557aa0d5de2552c3a

My reasoning behind this was this code part in the previous impl:
https://github.com/rubinius/rubinius/pull/3617/files#diff-b654709dae0e13be432ffb42c94a555eL454

We used the block scope after duping the proc but also we duped the block first.
Which is kinda confusing for me, so I probably need some confirmation if the new impl is doing the same (at least all specs still pass) but confirmation from @brixen, @YorickPeterse or anybody else who is confident enough about this stuff would be much appreciated.

brixen:
@tak1n this is a great first step, thanks for working on it. We'll continue to evolve some of these APIs to be even more simple. If you feel confident with this now, go ahead and merge, please! :+1:

PR #3621, opened 2016-02-17
Add issue template

jc00ke:
Seen at dear-github/dear-github#125

Opening as PR in case we want to discuss.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,27 +6,6 @@ performance improvements, and help with documentation. Every contribution is
 meaningful, so thank you for participating. That being said, here are a few
 guidelines that we ask you to follow so we can successfully address your issue.
 
-### Submitting Issues
-
-Please include the following:
-
-* The Rubinius version (`rbx -v`)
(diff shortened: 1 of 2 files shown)


chuckremes:
This is better than what we currently have, so I suggest just merging this. I'll merge it tomorrow if there aren't any strenuous objections before then.

YorickPeterse:
I wonder if people will actually use it though. A lot of times the "have you checked the contributing guide?" banner is ignored so I'm a bit sceptical about whether this would work or not.

The most important information one should provide (in this order) is:

1. A script/application (publicly available, we're not going to SSH into someone's production server) that reproduces the problem
2. The full contents of `config.rb`
3. The full contents of `configure.log`
4. Platform information (OS, distribution, etc)

I think the current template is a bit bare-bones as it is.

jc00ke:
My guess is that, because this pre-populates the issue input field, people will follow the format. They're def free to `Ctrl a del` if they want.

As for order, I agree. I'll rework it now.

chuckremes:
Let's merge it and iterate.

jc00ke:
:+1: thanks @chuckremes

Issue #3540, opened 2015-12-23
Keyword arguments sometimes are ignored, sometimes segfault


chanks:
On rbx 2.5.8:

```
chris@lovelace ~ $ irb
rbx-2.5.8 :001 > class Reproduction
rbx-2.5.8 :002?>     attr_reader :arg
rbx-2.5.8 :003?>   
rbx-2.5.8 :004 >       def initialize(arg:)
rbx-2.5.8 :005?>         @arg = arg
rbx-2.5.8 :006?>       end
rbx-2.5.8 :007?>   end
 => :initialize 
rbx-2.5.8 :008 > 
rbx-2.5.8 :009 >   10000.times do
rbx-2.5.8 :010 >       r = Reproduction.new(arg: 1)
rbx-2.5.8 :011?>     puts "yes!" if r.arg.nil?
rbx-2.5.8 :012?>   end
yes!
 => 10000 
```

So the instance variable isn't set every once in a while. If I push the loop even further I get:

```
chris@lovelace ~ $ irb
rbx-2.5.8 :001 > class Reproduction
rbx-2.5.8 :002?>     attr_reader :arg
rbx-2.5.8 :003?>   
rbx-2.5.8 :004 >       def initialize(arg:)
rbx-2.5.8 :005?>         @arg = arg
rbx-2.5.8 :006?>       end
rbx-2.5.8 :007?>   end
 => :initialize 
rbx-2.5.8 :008 > 
rbx-2.5.8 :009 >   100000.times do
rbx-2.5.8 :010 >       r = Reproduction.new(arg: 1)
rbx-2.5.8 :011?>     puts "yes!" if r.arg.nil?
rbx-2.5.8 :012?>   end
yes!
The Rubinius process is aborting with signal: SIGSEGV
--- begin system info ---
node info: lovelace #52~14.04.1-Ubuntu SMP Fri May 8 09:43:57 UTC 2015
--- end system info ---
--- begin rubinius info ---
process info: chris rbx 18285 2.5.8 2.1.0 2015-12-23 bef51ae3 3.4 JIT
--- end rubinius info ---
--- begin system backtrace ---
irb() [0x688f84]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x10340) [0x7ff6856b8340]
[...]

PR #3626, opened 2016-03-06
Protect self and CompiledCode from GC before handling arguments

ahmadsherif:
When arguments are keywords, `GenericArguments::call` will call `Rubinius::Runtime.keyword_object?`, which results in allocating some objects. When the young zone is full, garbage collection is run, which could lead to forwarding the self and/or the CompiledCode objects without updating their references, resulting in a faulty behavior or crashes like in #3540.

However, I'm not sure about the solution, so a feedback would be appreciated :smiley: .

### DIFF
--- a/vm/machine_code.cpp
+++ b/vm/machine_code.cpp
@@ -737,16 +737,23 @@ namespace rubinius {
 
       InterpreterCallFrame* frame = ALLOCA_CALLFRAME(mcode->stack_size);
 
-      // If argument handling fails..
-      if(ArgumentHandler::call(state, mcode, scope, args, previous) == false) {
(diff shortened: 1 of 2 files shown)


brixen:
@ahmadsherif thanks for looking into this. The situation you identified is a big problem across many parts of the codebase and is the reason I started on the work in the stw branch. That branch makes all allocation paths GC-safe (ie the GC will never run during allocation), among other things.

Since this is only one of several issues and I'm really close to merging stw, I'd like to hold off on these changes.

ahmadsherif:
@brixen No problem, thanks you for clarifying :smiley: .

PR #3635, opened 2016-04-06
Use sha256 for our homebrew formula

jsyeo:
This actually fixes https://github.com/rubinius/homebrew-apps/pull/3. @brixen, do you mind taking a look at the changes?


brixen:
@jsyeo I think it looks good, thanks!

PR #3639, opened 2016-04-24
Compile in debug mode when in git submodule

dylanmckay:
We currently compile in debug mode if there is a `.git` directory in the project
root.

This compiles Rubinius in debug mode even if `.git` is a file, so that
we compile in debug mode when rubinius is in a git submodule.

### DIFF
--- a/rakelib/release.rb
+++ b/rakelib/release.rb
@@ -3,6 +3,11 @@ def git_directory
   File.directory?(git_dir) && git_dir
 end
 
+def in_git?
+  git_dir = File.expand_path "../../.git", __FILE__
+  File.exist?(git_dir) && git_dir
+end
+
 def revision_file
   File.expand_path "../../.revision", __FILE__
 end
--- a/configure
+++ b/configure
@@ -167,7 +167,7 @@ class Configure
     @ruby_libversion = @ruby_version.split(/\./)[0..1].join.to_i
 
     # Configure settings
-    @release_build = !git_directory
+    @release_build = !in_git?
   end
 
   # Set up system commands to run in cmd.exe on Windows. Either Windows


rakelib/release.rb
```diff
@@ -3,6 +3,11 @@ def git_directory
   File.directory?(git_dir) && git_dir
 end
 
+def in_git?
+  git_dir = File.expand_path "../../.git", __FILE__
+  File.exist?(git_dir) && git_dir
```
YorickPeterse: Why is there the extra `&& git_dir` here? As far as I'm aware `File.expand_path` never returns nil.

brixen: @YorickPeterse I'm assuming it's just copied from above. The code in `git_directory` does that to pass the path back. A multi-clause conditional would have made that more clear. Since `#in_git?` looks like it's only intended to be a predicate, passing the path back doesn't appear needed, but it doesn't hurt either. Since this code will all be rewritten in bash soon-ish, it doesn't matter.

I'll merge and if you want to clean it up, feel free.

brixen:
@dylanmckay just to check my understanding on a couple points:

1. in a git submodule, the `.git` entity is a file instead of a directory?; and
2. the Rubinius build doesn't compile in debug mode unless you pass `--debug-build`, but if not in git, it assumes all the files are locally available instead of looking on the network. This is basically the gem files right now. Is this what you meant or did you mean something else?

dylanmckay:
@brixen 

1. Yes, it is a file
2. Yes, you're right, I misunderstood that but it makes sense now.

brixen:
@dylanmckay ok, cool, thanks!

PR #3651, opened 2016-05-05
Add JSON representation to those emitted by InstructionParser.

brixen:
This is intended to be a machine-readable description of all
VM instructions to be used to produce arbitrary forms of
documentation and tooling for the VM without requiring integration
into InstructionParser itself.

### DIFF
--- a/rakelib/instruction_parser.rb
+++ b/rakelib/instruction_parser.rb
@@ -1,3 +1,5 @@
+require 'json'
+
 # A simple parser to transform a description of the VM instructions into
 # various implementation files.
 #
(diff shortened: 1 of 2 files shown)


brixen:
All of the instruction data is now contained in .json files per instruction. See https://github.com/rubinius/rubinius/commit/2f599f8a86a20dec8f895a06113da15f81f8a7a4. There is still more work to do to generate all the compiler and VM files that are needed from those json files, but the new configuration addresses the use case for this ticket and branch.

/c @jemc

jemc:
Thanks for taking care of it - sorry I haven't been able to be of more help recently.

brixen:
@jemc no worries! Massive restructuring of the instructions to support all the new functionality (like functions), so I'll be curious to hear your thoughts when you have time. Working up a demo of compiling a stand-alone exe from a simple file with functions, so this may bring us pretty close to experimenting with Pony-style stuff.

PR #3656, opened 2016-05-20
Remove Ruby from the build system.

brixen:
The use of Ruby in the build sytem severely complicates building and packaging
Rubinius and has resulted in Rubinius either not getting packaged for various
platforms or even removed from existing packages like Homebrew.


brixen:
I've made a lot of progress on removing the Ruby dependency, but it's not done yet. Closing this because these parts are implemented.

PR #3660, opened 2016-06-05
Enable ccache for faster builds

thedrow:
Caching is now enabled on all environments.
It should speed up the build significantly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+cache: ccache
 sudo: required
 dist: trusty
 compiler:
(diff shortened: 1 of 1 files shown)


thedrow:
Looks like that the build system doesn't use the path. I think it uses CC and CXX instead.
How do we resolve that?

thedrow:
So I got around the build system not recognizing ccache but there's a spec failure on linux now.
Any idea why?

brixen:
@thedrow this build failure on Linux is most likely the issue related to finalization or finalizer thread synchronization that I'm working on so I'm going to merge this.

Taher-Ghaleb:
Hello,

Sorry, I know it's been a long time ago, but I am curious why caching hasn't been enabled before this (i.e., since December 17, 2014, when it became available to open source projects).

Thank you.

PR #3681, opened 2016-07-22
Add a failing spec for a case then protected method called from public_send in own method

kachick:
ref: https://bugs.ruby-lang.org/issues/7499


brixen:
@kachick thanks for the patch! It's a small thing, but could you please use `lambda` in place of `->` so the specs are consistent.

kachick:
@brixen Sorry, I forgot it 🙇

Adjusted in https://github.com/rubinius/rubinius/pull/3681/commits/10214f2a9083ed2513e810d3a7c19eeab9264220, Thank you!

brixen:
@kachick thank you!

kachick:
😄

PR #3695, opened 2016-09-17
Update libsodium to 1.0.11

jc00ke:
Bump from 1.0.3


jc00ke:
img_20160917_130330

Addie's first PR!

brixen:
Thanks, @jc00ke and Addie! :) Does she have a GitHub account yet?

RE libtool, curious which problem you are trying to solve:

1. auto-installing dependencies
2. ensuring they are up-to-date
3. both
4. neither

For *updating* libtool under Homebrew, this is one possible option:

```
brew outdated libtool || brew upgrade libtool
```

If libtool is not already installed, we could probably combine commands like this

```
brew list libtool || brew install libtool
brew outdated libtool || brew upgrade libtool
```

Generally, we don't auto-install dependencies right now, so doing that would be a departure from what we do now. I don't object to that, and planned to make auto-installing dependencies a component of the non-Ruby build scripts in the future.

If we do it now, I'd suggest formalizing it just a bit in `configure` with a section that explicitly calls out probing and installing. It doesn't need to be complete, but a bit of initial structure makes a place people can continue to extend as we move forward.

jc00ke:
I need to get her an email account first, and some `ssh` & `gpg` keys :wink: 

I'll choose door 4, neither. It seems as though `libtool` is already installed ([line 814](https://travis-ci.org/rubinius/rubinius/jobs/160711538#L814-L817) shows its use and then how the `libsodium` `autogen` script can't find it) but unable to be called. I borrowed Theresa's Mac and was able to get `libsodium` to build only after installing `libtool` via `homebrew` and having it add the `g` in front, as to not conflict with the previously installed version.

I'm not sure if _upgrading_ `libtool` will work, as I think it's a linking problem, but I'm unfamiliar with how `homebrew` and `OS X` work in this regard.

Somewhere between `libsodium` 1.0.3 and 1.0.11 the `./autogen.sh` step was added, at least when pulled directly from the source.

Now that I'm looking at [releases](https://download.libsodium.org/libsodium/releases/) I see that those archives already have the `configure` script, so I can pull from one of those and can (potentially) back out the `libtool` stuff.

jc00ke:
Closing in favor of #3696.

Issue #3723, opened 2017-01-08
Array#sample doesn't seem to sampling correctly


tompng:
The result of `[0,1,2,3].sample(3).first` seems to be biased.

I checked the randomness of Array#sample by running this code
```ruby
p 10000.times.map{[0,1,2,3].sample(3)[0]}.group_by{|a|a}.map{|a,b|[a,b.size]}.sort_by(&:first)
```
got `[[0, 2415], [1, 3333], [2, 2867], [3, 1385]]`
expected result is
`[[0, 2300~2700], [1, 2300~2700], [2, 2300~2700], [3, 2300~2700]]`

<hr>
1. What command did you run?

```sh
my-shell$ docker pull rubinius/docker
my-shell$ docker run -it rubinius/docker bash
docker-shell# rbx -v
rubinius 3.69 (2.3.1 a57071c6 2016-11-17 3.8.0) [x86_64-linux-gnu]
docker-shell# rbx -e "p 10000.times.map{[0,1,2,3].sample(3)[0]}.group_by{|a|a}.map{|a,b|[a,b.size]}.sort_by(&:first)"
[[0, 2415], [1, 3333], [2, 2867], [3, 1385]]
```

2. What behavior did you expect?
```ruby
% ruby -v
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin16]
% ruby -e "p 10000.times.map{[0,1,2,3].sample(3)[0]}.group_by{|a|a}.map{|a,b|[a,b.size]}.sort_by(&:first)"
[[0, 2490], [1, 2585], [2, 2473], [3, 2452]] # each should be 2300~2700
```

4. What version of Rubinius?
  Output of `rbx -v`: `rubinius 3.69 (2.3.1 a57071c6 2016-11-17 3.8.0) [x86_64-linux-gnu]`

5. What version of operating system?

  Output of `uname -a`: `Linux fc6d004976bf 4.4.39-moby #1 SMP Fri Dec 16 07:34:12 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`

brixen:
Fixed by #3724.

PR #3724, opened 2017-01-09
fix Array#sample randomness

tompng:
Fixes issue #3723

<hr>

  - [x] It fixes an issue with an existing features.
  - [ ] It introduces a new feature.

  - [x] Yes, it includes tests.

### DIFF
--- a/core/array.rb
+++ b/core/array.rb
@@ -1516,30 +1516,13 @@ class Array
         while i < count
           k = rng.rand(size)
 
-          spin = false
-          spin_count = 0
-
-          while true
-            j = 0
-            while j < i
-              if k == result[j]
-                spin = true
-                break
-              end
-
-              j += 1
-            end
-
-            if spin
-              if (spin_count += 1) > 100
-                abandon = true
-                break
-              end
-
-              k = rng.rand(size)
-            else
+          j = 0
+          while j < i
+            if k == result[j]
+              abandon = true
               break
             end
+            j += 1
           end
 
           break if abandon
--- a/core/array.rb
+++ b/core/array.rb
@@ -1563,7 +1563,7 @@ class Array
       result = Array.new self
       tuple = Rubinius::Mirror::Array.reflect(result).tuple
 
-      count.times { |i| tuple.swap i, rng.rand(size) }
+      count.times { |i| tuple.swap i, rng.rand(size - i) + i }
 
       return count == size ? result : result[0, count]
     end
--- a/spec/ruby/core/array/sample_spec.rb
+++ b/spec/ruby/core/array/sample_spec.rb
@@ -2,6 +2,13 @@ require File.expand_path('../../../spec_helper', __FILE__)
 require File.expand_path('../fixtures/classes', __FILE__)
 
 describe "Array#sample" do
+  it "samples evenly" do
+    3.times do |i|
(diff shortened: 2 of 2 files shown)


core/array.rb
```diff
@@ -1516,30 +1516,13 @@ def sample(count=undefined, options=undefined)
         while i < count
           k = rng.rand(size)
 
-          spin = false
-          spin_count = 0
-
-          while true
```
tompng: Once `spin` is set to true, `spin` will never become false in this loop.
In that case, `spin_count` will always becomes over 100, and always abandoned.
(so I removed `spin`, `spin_count`, and this `while true` loop)

core/array.rb
```diff
@@ -1516,30 +1516,13 @@ def sample(count=undefined, options=undefined)
...
-            while j < i
-              if k == result[j]
-                spin = true
-                break
-              end
-
-              j += 1
-            end
-
-            if spin
```
tompng: If you add spin=false here(original algorithm with no bug?), the calculation cost might be very large.

core/array.rb
```diff
@@ -1563,7 +1546,7 @@ def sample(count=undefined, options=undefined)
       result = Array.new self
       tuple = Rubinius::Mirror::Array.reflect(result).tuple
 
-      count.times { |i| tuple.swap i, rng.rand(size) }
+      count.times { |i| tuple.swap i, rng.rand(size - i) + i }
```
tompng:
```
https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
Implementation errors
Similarly, always selecting j from the entire range of valid array indices on every iteration also produces a result which is biased, albeit less obviously so
```

spec/ruby/core/array/sample_spec.rb
```diff
@@ -2,6 +2,13 @@
 require File.expand_path('../fixtures/classes', __FILE__)
 
 describe "Array#sample" do
+  it "samples evenly" do
+    3.times do |i|
+      samples = 4000.times.map { 4.times.to_a.sample(3)[i] }
+      counts = samples.group_by(&:itself).map { |a, b| b.size }
+      counts.each { |count| (800..1200).should include(count) }
```
tompng: How I choose 800..1200:
probability of count in 800..1200 is 0.9999999999996083

if you change this to 900..1100, probability will be 0.99976
it means that the test wil fail in 0.024% even if there is no bug.

```ruby
def overflow_safe_mult values
  a = 1
  values = values.sort_by(&:abs)
  a *= a.abs>1 ? values.shift : values.pop until values.empty?
  a
end
def probability_of_count n, p, range
  range.map{|i|
    overflow_safe_mult [p]*i+[1-p]*(n-i)+i.times.map{|j|n-j}+i.times.map{|j|1.0/(j+1)}
  }.inject(:+)
end
probability_of_count 4000, 0.25, 0..799 #=> 3.181654463145792e-14
probability_of_count 4000, 0.25, 800..1200 #=> 0.9999999999996083
probability_of_count 4000, 0.25, 1201..4000 #=> 3.572627145186968e-13

probability_of_count 4000, 0.25, 900..1100 #=> 0.999756654675964
```

brixen:
@tompng thank you for the PR and for the detailed explanations, much appreciated! ❤️

Issue #3296, opened 2015-01-27
Random::DEFAULT is missing


godfat:
MRI/JRuby has this, and unfortunately, it does not seem to make much sense to have this in Rubinius given how Rubinius implements random. At first I just want to assign it with `Thread.current.randomizer` or `Thread.main.randomizer`, but a `Rubinius::Randomizer` is not a `Random`.

Any idea how we could add this, or just don't define it? I could probably just use `Random.seed` in my code, but MRI does not provide this method. Perhaps we should just ask MRI implements `Random.seed` instead?

godfat:
Oops, I misread the code. There's no `Random.seed` defined in Rubinius.

brixen:
@godfat I don't see a problem adding this constant. AFAIK, it's just an instance of Random. Is there some reason you see adding it would be a problem?

godfat:
@brixen Since `Rubinius::Randomizer` is stored inside a thread local, I would guess it's not thread safe. Setting `Random::DEFAULT` would mean it's shared amongst all the threads. Also, what `Rubinius::Randomizer` should the `Random` instance refer to? The one from main thread, or a whole new one? This might also take some memory which most people might never use.

eugenk:
Is there any news on this? We are using `Random::DEFAULT` indirectly via an external gem in our tests now and Rubinius not having this constant makes the tests fail.

eugenk:
I've worked around it in our test suite, but the question remains: Any news on this?

godfat:
Sorry, I don't have anything new to share.

brixen:
@godfat @eugenk I don't see any evidence that MRI synchronizes calls to the object that `Random::DEFAULT` points to so Ruby implicitly doesn't make any concurrency guarantees about this object. This is similar to an Array instance or any number of other Ruby core library classes.

The simplest solution for Rubinius is probably just to assign an instance of Random to `Random::DEFAULT`. Could someone open a PR for this?

PR #3752, opened 2017-07-25
Add Random::DEFAULT as an instance of Random

godfat:
This is basically for MRI compatibility. This could be verified by:

    > ruby -ve 'p ObjectSpace.each_object(Random).to_a == [Random::DEFAULT]'
    ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
    true

Closes https://github.com/rubinius/rubinius/issues/3296

  - [ ] It fixes an issue with an existing features.
  - [x] It introduces a new feature.

  - [x] Yes, it includes tests.

Unfortunately sorry that I have some difficulty to get Rubinius HEAD running, therefore I am not sure if this would work. One of the issues I have is:

```
Unable to send 'object_equal' on '#<Module>' (#<Class>)


0x7f92c1d1e170: Kernel#raise in core/alpha.rb:146 (+56)
0x7f92c1d1e800: Kernel#object_equal in core/alpha.rb:172 (+29)
0x7f92c1d1eeb0: BasicObject#! in core/basic_object.rb:15 (+6)
0x7f92c1d1f500: Random#initialize in core/random.rb:148 (+41)
0x7f92c1d1fb70: Class#new in core/alpha.rb:93 (+16)
0x7f92c1d20200: Object#script in core/random.rb:186 (+55)
```

### DIFF
--- a/core/zed.rb
+++ b/core/zed.rb
@@ -1853,6 +1853,10 @@ class Time
   }
 end
 
+class Random
+  DEFAULT = new
+end
+
 module Rubinius
   class CodeDB
     class << self
--- a/spec/ruby/core/random/default_spec.rb
+++ b/spec/ruby/core/random/default_spec.rb
@@ -1,0 +1,7 @@
+require File.expand_path('../../../spec_helper', __FILE__)
+
+describe "Random::DEFAULT" do
+  it "returns a Random instance" do
+    Random::DEFAULT.should be_an_instance_of(Random)
+  end
+end


godfat:
Oh, I just realized that this is related to my patch:

    0x7f92c1d1f500: Random#initialize in core/random.rb:148 (+41)

So we should probably initialize that later.

core/random.rb
```diff
@@ -182,3 +182,5 @@ def bytes(length)
     s
   end
 end
+
+Random::DEFAULT = Random.new
```
brixen: I'd suggest adding this line to the end of core/zed.rb. The Rubinius core libraries load in a sort of phased process where some functionality isn't ready when a particular file is loaded. The core/zed.rb file is loaded last and all basic Ruby functionality should be available at that time, but some pieces of the core library still need to be assembled.

brixen:
@godfat thanks for working on the PR! With my suggested change above, I'm able to build locally.

core/random.rb
godfat: Oh, that's a smart trick. I was looking for directories like `delta` but I guess we're no longer doing that.

Completely off topic: May I ask if `zed.rb` means anything? If it's just lexically sorted the last, I would imagine using something like `zzz.rb` would be funny.

godfat:
@brixen Thanks for the review! I amended and the test would also pass locally for me.

brixen:
@godfat thanks!

Issue #3295, opened 2015-01-24
Keyword arguments are not passed to parent class when calling super without arguments


jirutka:
#### Example

``` ruby
class Panthera
  def initialize(name: 'Panthera')
    @name = name
  end

  def to_s
    "name: '#{@name}'"
  end
end

class Tigris < Panthera
  def initialize(name: 'Tiger')
    super  # <-- HERE!
  end
end

puts Tigris.new(name: 'My Tiger')
puts Tigris.new
```
#### Expected output:

```
name: 'My Tiger'
name: 'Tiger'
```

Tested on MRI 2.1.2, MRI 2.2.0, JRuby 9.0.0.0.pre1.
#### Actual output:

```
name: ''
An exception occurred running script.rb:

method 'initialize': given 1, expected 1 (ArgumentError)

Backtrace:

                 Tigris#initialize at script.rb:13
                 Object#__script__ at script.rb:18
  Rubinius::CodeLoader#load_script at kernel/delta/code_loader.rb:66
  Rubinius::CodeLoader.load_script at kernel/delta/code_loader.rb:152
           Rubinius::Loader#script at kernel/loader.rb:645
             Rubinius::Loader#main at kernel/loader.rb:799
```

Tested on rubinius 2.5.0 (2.1.0 50777f41 2015-01-17 3.5.1 JI) [x86_64-darwin13.4.0].

brixen:
@jirutka thanks for reporting this. It should be a ridiculous amount of fun figuring out all the edge-cases here, like how a keyword object is split into separate Hashes based on the types of keys. Any chance you'd be able to add some specs for this behavior? See eg https://github.com/rubinius/rubinius/blob/master/spec/ruby/language/method_spec.rb and https://github.com/rubinius/rubinius/blob/master/spec/ruby/language/block_spec.rb

andrew-aladev:
```
class A
  def fit q = 1, w = 1, a: 1
    puts a
  end
end

class B < A
  def fit q = 1, w = 1, a: 1
    super q, w, a: a
  end
end

class C < A
  def fit q = 1, w = 1, a: 1
    super q, w, a
  end
end

class D < A
  def fit q = 1, w = 1, a: 1
    super
  end
end

B.new.fit
C.new.fit
D.new.fit
```

I've debuged `GenericArguments::call` a bit. We are failing here:
```
// Too many arguments
if(mcode->keywords && !RP && !KP && E > O) return false;
```

```
B.new.fit:
[...]

PR #3758, opened 2017-09-27
Fixed keyword and placeholder arguments in zsuper instruction, improved tests

andrew-aladev:
Today I've finished issue #3295: `zsuper` instruction code was a bit broken. I've fixed it and improved tests for `super` method.

I left a `TODO` that `keywrods`wants to have `machine/class/hash` available. This class can improve readability of this code.

Existing spec was very poor. I've implemented a test that uses all combinations of 3 arguments types: regular, keywords and placeholder.

Please try to run improved tests without fix of `zsuper` instruction. You can see that all argument types are affected by this issue.

```
The super keyword when using keyword arguments does not pass any arguments to the parent when none are given ERROR
ArgumentError: method 'foo': given 1, expected 1

The super keyword when using keyword arguments passes default argument values to the parent ERROR
NoMethodError: undefined method `[]=' on nil:NilClass.

The super keyword when using regular and keyword arguments passes only required regular arguments to the parent when no 
optional keyword arguments are given ERROR
ArgumentError: method 'foo': given 2, expected 2

The super keyword when using regular and keyword arguments passes default argument values to the parent ERROR
ArgumentError: method 'foo': given 2, expected 2

The super keyword when using regular and keyword arguments passes any given regular and keyword arguments including optional 
keyword arguments to the parent FAILED
Expected ["a", {:b=>"b"}]
 to equal ["a", {:b=>"b", :c=>"c"}]

The super keyword when using splat and keyword arguments does not pass any arguments to the parent when none are given FAILED
Expected [[nil], {}]
 to equal [[], {}]

The super keyword when using splat and keyword arguments passes only splat arguments to the parent when no keyword arguments 
are given FAILED
Expected [["a", nil], {}]
```

### DIFF
--- a/machine/instructions/zsuper.hpp
+++ b/machine/instructions/zsuper.hpp
@@ -62,19 +62,38 @@ namespace rubinius {
 
       if(mc->keywords) {
         native_int placeholder_position = splat_obj ? mc->total_args : mc->total_args - 1;
-        native_int keywords_position = placeholder_position + 1;
+        native_int keywords_position    = placeholder_position + 1;
 
-        Object* placeholder = scope->get_local(state, placeholder_position);
-        Array* ary = Array::create(state, 2);
+        // "Runtime.keyword_object?" requires keywords object to be convertable into Hash.
+        // TODO define "machine/class/hash" and use Hash::create instead of "[[key, value], ..].to_h".
 
-        for(native_int i = keywords_position; i <= mc->keywords_count; i++) {
-          ary->set(state, 0, as<Symbol>(call_frame->compiled_code->local_names()->at(state, i)));
-          ary->set(state, 1, scope->get_local(state, i));
+        Array* keywords = Array::create(state, mc->keywords_count);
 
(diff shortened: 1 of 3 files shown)


brixen:
@andrew-aladev this is awesome! Thank you for working on it.

There are definitely things that will be improved in the instructions, including new ways to set up the arguments for method calls, so this will move us forward by fixing those existing issues. We won't be adding `machine/class/hash` and we'll be removing most-to-all of what is in `machine/class` right now as we build out the ability to specify C structs via LLVM directly from the instruction set. The goal is to reduce the C++ code to the essential minimum in the interpreter, garbage collector, JIT, and underlying concurrency mechanism. All this isn't necessarily relevant to your patch, but I wanted to give you an idea of future work.

Fixing this issue is a big step forward. Thanks again for working on it.

chuckremes:
@andrew-aladev Thanks for this patch! It fixes a problem that has been irritating me for a while. Really appreciate your fix.

Issue #3738, opened 2017-03-18
Enumerable#each_entry behaves differently from MRI


toy:
`Enumerable#each_entry` yields different from MRI value for no arguments.

Ran using docker image, `ruby -v`:
rubinius 3.72 (2.3.1 6d56979c 2017-02-21 3.8.0) [x86_64-linux-gnu]

test file:
```rb
class Foo
  include Enumerable

  def each
    yield 1
    yield 1, 2
    yield
  end
end

out = []
Foo.new.each_entry{ |*args| out << args }
puts "each_entry: #{out.inspect}"
```

Expected output:
```
==[1.9.3-p551]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.0.0-p648]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.1.10]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.2.6]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.3.3]==
each_entry: [[1], [[1, 2]], [nil]]

==[2.4.0]==
each_entry: [[1], [[1, 2]], [nil]]
```
Rubinius output:
```
each_entry: [[1], [[1, 2]], [[]]]
```

I suggest following change:
```diff
diff --git core/enumerable.rb core/enumerable.rb
index 2234f9b61d..a35e2eb1ff 100644
--- core/enumerable.rb
+++ core/enumerable.rb
@@ -67,8 +67,8 @@ module Enumerable

   def each_entry(*pass)
     return to_enum(:each_entry, *pass) { enumerator_size } unless block_given?
-    each(*pass) do |*element|
-      yield element.size == 1 ? element[0] : element
+    each(*pass) do |*elements|
+      yield elements.size > 1 ? elements : elements[0]
     end
     self
   end
```

Related: jruby/jruby#4532

kachick:
@toy Thanks for your reporting! Sorry to late response 🙇 We fixed the problem in https://github.com/rubinius/rubinius/pull/3761.

PR #3761, opened 2017-11-04
Fix Enumerable#each_entry with no yield arguments

kachick:
Fix https://github.com/rubinius/rubinius/issues/3738


brixen:
@kachick thanks!

PR #3766, opened 2017-11-04
Add a failing spec for Enumerator#peek with multiple yield arguments

kachick:
I don't have simple idea to fix it 🤔  So just adding the failing specs for now.

### DIFF
--- a/spec/ruby/core/enumerator/peek_spec.rb
+++ b/spec/ruby/core/enumerator/peek_spec.rb
@@ -33,4 +33,19 @@ describe "Enumerator#peek" do
     5.times { @e.next }
     lambda { @e.peek }.should raise_error(StopIteration)
   end
+
+  context "when yielded with multiple arguments" do
+    before do
+      object = Object.new
+      def object.each
+        yield 1, 2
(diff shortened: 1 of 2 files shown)


brixen:
@kachick thanks for adding these. So, I haven't dug into this, but a couple things come to mind: 1. we don't need to support non-Ruby-2.x behavior, so may be some remnants of old Ruby block argument processing that may linger and can be removed; 2. my recollection is fuzzy on it now, but I think there were some Ruby core methods that did some block argument tricks to mimic old behavior. If 2 is correct, we may want to standardize an approach to this in the compiler, instead of invoking a special instruction in the Ruby code.

Does this track at all what you've been seeing, or am I trapped in a bad lucid nightmare about the 1.8 to 1.9+ Ruby semantics migration?

brixen:
@kachick thanks! I'll figure out my questions in the comment above. Working on redoing method invocation, which will probably touch block argument processing too.

kachick:
Thank you for the merging!
(Sorry to late response, I overlooked. :bow:)

PR #3770, opened 2017-11-15
Fix `Regexp.union` behavior on optional left parenthesis

akhramov:
`Regexp::SourceParcer` always treats left parenthesis as a beginning of
a group. However, the parenthesis can be escaped and in this case an error will be raised.

Reproduction code:

```ruby
Regexp.union(/\(?/, "")
```

This change tweaks the `Regexp::SourceParcer#create_parts` method to
handle this case.

  - [x] It fixes an issue with an existing features.
  - [ ] It introduces a new feature.

  - [x] Yes, it includes tests.

### DIFF
--- a/core/regexp.rb
+++ b/core/regexp.rb
@@ -388,8 +388,10 @@ class Regexp
     def create_parts
       while @index < @source.size
         if @source[@index].chr == '('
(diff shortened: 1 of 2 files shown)


brixen:
@akhramov thank you!

PR #3791, opened 2018-03-30
Update README

eregon:
* Remove the mention of a JIT compiler since it was removed in e4cbba00.

### DIFF
--- a/README
+++ b/README
@@ -3,9 +3,9 @@
 Rubinius is a modern language platform that supports a number of programming
 languages.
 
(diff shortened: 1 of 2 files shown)


sempervictus:
The JIT should be coming back, in a format grokable and extensible by us mortals... Without it, rbx is almost unusable in terms of its performance for production workloads (things that work on a budget and thus at least partially quantified in a perf/$ manner, along with function, stability, and security).

brixen:
The JIT is being worked on. It'll always be in some state of progress.

PR #3817, opened 2019-06-06
Clean up timestamp files generated by Makefile

ljulliar:
core clean task must delete timestamp files created by make while building Rubinius or the next build operation relaunched after a rake clean will fail


brixen:
Thanks!

Issue #3755, opened 2017-09-07
Can't install Rubinius using RVM on mac


adamthedeveloper:
Thank you for taking the time to let us know that you're having an issue with Rubinius. We apologize for the inconvenience.

Please respond to the following questions to help us fix the issue. Please leave the form contents in place. If a question isn't relevant, please respond with N/A.

Please include a link to a gist (https://gist.github.com/) for terminal output, including backtraces, rather than pasting them inline in the issue.

Here you go: [https://gist.github.com/adamthedeveloper/7432ec4970499ca7436c0c1cd962833d](https://gist.github.com/adamthedeveloper/7432ec4970499ca7436c0c1cd962833d)

1. What command did you run?

  I ran `rvm install rbx`
  I also tried `rvm install rubinius`

2. What behavior did you expect?

  I expected it to succeed just like MRI versions succeed.

3. What behavior did you get instead?

Here's an excerpt of the (unhelpful) output (llnk to full log above): 

checking for assert.h... yes
checking for openssl/ssl.h... yes
checking for OpenSSL_add_all_digests() in -lcrypto... no
checking for OpenSSL_add_all_digests() in -llibeay32... no
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.

Provided configuration options:
	--with-opt-dir
	--without-opt-dir
	--with-opt-include
	--without-opt-include=${opt-dir}/include
	--with-opt-lib
	--without-opt-lib=${opt-dir}/lib
	--with-make-prog
	--without-make-prog
	--srcdir=.
	--curdir
	--ruby=/Users/adammedeiros/.rvm/src/rbx-3.84/staging/bin/rbx
	--with-openssl-dir
	--without-openssl-dir
	--with-openssl-include
	--without-openssl-include=${openssl-dir}/include
	--with-openssl-lib
	--without-openssl-lib=${openssl-dir}/lib
	--with-kerberos-dir
	--without-kerberos-dir
	--with-kerberos-include
	--without-kerberos-include=${kerberos-dir}/include
	--with-kerberos-lib
	--without-kerberos-lib=${kerberos-dir}/lib
	--with-debug
	--without-debug
	--enable-debug
	--disable-debug
	--with-nsllib
	--without-nsllib
	--with-socketlib
	--without-socketlib
	--with-openssl-config
	--without-openssl-config
	--with-pkg-config
[...]

PR #3826, opened 2019-09-12
Add installation note about openssl.

JonRowe:
Having tried to install Rubinius on OS X I ran into the problem from #3755 and it took me a heap of searching to find the issue, how about adding a note to the read me explaining this caveat?


brixen:
@JonRowe thanks!

PR #3848, opened 2022-03-29
configure.rb: Add alternate gettid() header (`unistd.h`)

### LINKED ISSUE #3847: Error 'gettid' is missing exception specification 'throw()' when compiling
On my system, attempting to compile the project will cause the following error:

    'gettid' is missing exception specification 'throw()'

This is caused by the fact that configure.rb is not detecting `gettid` as present, causing a conflict with the current `gettid` glibc definition (Rubinius doesn't have a noexcept specifier, while gblic's version does).

I'm submitting a PR shortly to fix this.

64kramsystem:
On Linux systems providing gettid(), the header is `unistd.h`, which is currently not evaluated. The consequence of this omission is that Rubinius' gettid() definition conflicts with Glibc's one - the former doesn't have the noexcept specifier, while the latter does (__THROW).

Closes #3847.

Extra context: this was not detected before because gettid() has been introduced on Glibc 2.30, which is not very old.

### DIFF
--- a/configure.rb
+++ b/configure.rb
@@ -1295,7 +1295,9 @@ int main() { return tgetnum(""); }
       @defines << "HAVE_INOTIFY"
     end
 
-    if has_function("gettid", ["unistd.d", "sys/types.h"])
+    # On Linux systems providing gettid() (Glibc 2.30+), the header is `unistd.h`.
+    #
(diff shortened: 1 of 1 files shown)


64kramsystem:
Closing for inactivity.

